PR #8, opened 2024-11-19
feat: check verification result for contradictions

p0deje:
Sometimes LLMs would incorrectly fail the verfication and provide an explanation that contradicts itself. For example, given the following verification "square A is positioned to the left of square B", LLM sometimes fails it but provide weird explanation: "The visual representation shows square A to the right of square B, indicating that square B is not positioned to the left of square A."

This commit compensates for such behavior by double-checking the verification explanation for contradictions.

### DIFF
--- a/alumnium/alumni.py
+++ b/alumnium/alumni.py
@@ -8,7 +8,7 @@ from langchain_google_genai import ChatGoogleGenerativeAI
 from retry import retry
 from selenium.webdriver.remote.webdriver import WebDriver
 
-from .agents import ActorAgent, VerifierAgent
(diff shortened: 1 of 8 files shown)


alumnium/alumni.py
```diff
@@ -8,15 +8,20 @@
...
 logger = logging.getLogger(__name__)
 
 
 class Alumni:
-    def __init__(self, driver: WebDriver, model: Model = Model.load()):
+    def __init__(
+        self,
+        driver: WebDriver,
+        model: Model = Model.load(),
+        check_contradictions: bool = True,
```
sh3pik, line 23: we can add it to the readme later. if it works 😄

PR #20, opened 2024-12-27
feat!: do not pass ARIA in vision checks

p0deje:


p0deje:
Closing in favor of #21

PR #21, opened 2024-12-29
Multiple fixes and improvements

p0deje:
Please see individual commits for details.

### DIFF
--- a/alumnium/tools/select_tool.py
+++ b/alumnium/tools/select_tool.py
@@ -3,7 +3,7 @@ from alumnium.drivers import SeleniumDriver
 
 
 class SelectTool(BaseModel):
-    """Selects an option in a dropdown."""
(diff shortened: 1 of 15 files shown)


alumnium/tools/select_tool.py
```diff
@@ -3,7 +3,7 @@
 
 
 class SelectTool(BaseModel):
-    """Selects an option in a dropdown."""
+    """Selects an option in a dropdown. Only use this tool if the dropdown is a combobox."""
```
sh3pik, line 6: this is the instruction for LLM, right? maybe better to start with `if`?
```suggestion
    """If a dropdown is a combobox, selects an option in the dropdown"""
```

p0deje: As far as I understand, the instructions should be written imperatively, i.e. "do this if that".

PR #24, opened 2025-01-25
feat: waits for requests and DOM before taking ARIA

sh3pik:
- removes Loading detector agent
- adds JS scripts for Selenium and Playwright to wait for DOM changes and requests
- disables the retry test. The loader on the page is a GIF, so the script can't detect this "changing".


alumnium/agents/verifier_agent.py
```diff
@@ -78,11 +74,4 @@ def invoke(self, statement: str, vision: bool = False) -> Verification:
         logger.info(f"  <- Reason: {verification.explanation}")
         logger.info(f'  <- Usage: {message["raw"].usage_metadata}')
 
-        if not verification.result:
-            loading = self.loading_detector_agent.invoke(aria, title, url, screenshot)
-            if loading and self.retry_count > 0:
-                sleep(LoadingDetectorAgent.delay)
-                self.retry_count -= 1
-                return self.invoke(statement, vision)
```
sh3pik, line 86: @p0deje I'm not sure about this change. I'm thinking of keeping it like this. wdyt?
```python
        if not verification.result and self.retry_count > 0:
                sleep(delay)
                self.retry_count -= 1
                return self.invoke(statement, vision)
```

PR #26, opened 2025-01-29
feat: replace verifier with retriever

p0deje:
What started initially as an attempt to implement 2-pass verification to increase its stability ended up being a complete replacement of `verifier/confirmation_checker` with `retriever/extractor` agents. For now it seems to work good enough, though I would like to further improve the implementations by utilizing chain-of-thought approach for structured outputs. This can be done in follow-up PRs.

Overall, the new design for verification is to:
1. Use retriever to extract "Is the following statement true or false - ..." from the screenshot/ARIA.
2. Use extractor to get `true/false` output of the retriever result.
3. Assert on the extracted value.

-----------------

**Smaller additions in this PR:**
- New variable `ALUMNIUM_LOG_LEVEL` which compliments existing `ALUMNIUM_DEBUG`.
- Playwright in tests now run in headless mode.
- Automatic retries of random AWS Bedrock errors which are frequent on Llama.
- Timeout CI jobs after 30 minutes.
- List separator changed from `<sep` to `%SEP%` because Haiku treats the former as opening HTML tag and likes to append the closing tag (e.g. `<sep>FOO</sep>`).


alumnium/alumni.py
```diff
@@ -75,28 +73,25 @@ def do(self, goal: str):
...
 
-    def check(self, statement: str, vision: bool = False) -> Verification:
+    def check(self, statement: str, vision: bool = False) -> str:
         """
         Checks a given statement using the verifier.
 
         Args:
             statement: The statement to be checked.
             vision: A flag indicating whether to use a vision-based verification via a screenshot. Defaults to False.
+            retries: The number of retries to check the statement. Defaults to the value set in the LoadingDetectorAgent.
```
sh3pik: @p0deje hm, there is no `LoadingDetectorAgent` anymore, right?

alumnium/__init__.py
```diff
@@ -1,11 +1,17 @@
...
 
-if os.getenv("ALUMNIUM_DEBUG", "0") == "1":
-    logger = logging.getLogger(__name__)
-    logger.setLevel(level=logging.DEBUG)
-    logger.addHandler(logging.StreamHandler(sys.stdout))
+
+logger = logging.getLogger(__name__)
+
+level = getenv("ALUMNIUM_LOG_LEVEL", None)
+if getenv("ALUMNIUM_DEBUG", "0") == "1":
```
sh3pik: Is it kept for backward compatibility?

p0deje: Not really, I can remove it if you think it's unuseful.

alumnium/alumni.py
p0deje: I'll remove

alumnium/__init__.py
sh3pik: I think it might be more confusing than helpful. 

I imagine the situation when `ALUMNIUM_DEBUG=1` and `ALUMNIUM_LOG_LEVEL=info`.  If I look at them in the `.env` file, I can forget which one gets a priority

PR #36, opened 2025-04-04
feat: improve determinism with top_p=0

p0deje:


alumnium/alumni.py
```diff
@@ -33,24 +33,26 @@ def __init__(self, driver: Page | WebDriver, model: Model = Model.load()):
                 model=model.value,
                 api_version=getenv("AZURE_OPENAI_API_VERSION", ""),
                 temperature=0,
+                top_p=1,
```
sh3pik: @p0deje shouldn't this be lower instead? like, 0.3? I might not fully understand what top_p means, though.

p0deje: @sh3pik This is what Gemini advises me to do:

> Ensure top_p is Not Interfering:
> top_p (nucleus sampling) is another parameter that controls randomness. It selects tokens from the smallest possible set whose cumulative probability exceeds p.
While temperature=0 should theoretically make top_p irrelevant (as there's only one highest-probability token), it's best practice to explicitly set top_p to a value like 1 when using temperature=0. This ensures nucleus sampling doesn't accidentally get applied or interfere in some edge cases or different model implementations.
Recommendation: Set temperature=0 and top_p=1.

alumnium/alumni.py
```diff
@@ -33,24 +33,26 @@ def __init__(self, driver: Page | WebDriver, model: Model = Model.load()):
                 model=model.value,
                 api_version=getenv("AZURE_OPENAI_API_VERSION", ""),
                 temperature=0,
+                top_p=0.3,
```
p0deje: Can we set it all the way down to 0?

sh3pik: @p0deje I'm unsure. let me try 0.1

p0deje: @sh3pik What's the lowest we can set? Have you tried 0?

sh3pik: @p0deje no, but the stability degraded on 0.1 already. I want to revert the latest commit. Let me try 0 JIC.

p0deje: @sh3pik Well, maybe it didn't make things worse, it just revealed the underlying problem in our prompts that we had not seen before. If it fails _consistently and all the time_, I'm happy to merge it and fix the prompts.

sh3pik: 🚀 let's do it then!

sh3pik (approved):

alumnium/alumni.py
p0deje: @sh3pik Can you check that it fails consistently with top_p=0 as well? I am not 100% sure, I was just thinking it's possible. I've read through a bunch of posts on OpenAI forums and Reddit and still don't know which value we should set (0 or a small non-zero value like 0.01). I feel like we need to experiment and evaluate different values for determinism.

Issue #58, opened 2025-04-12
Run format job on CI


p0deje:
We have the `make format` command that can be used to run various linters and we need a CI job in `ci.yml` that is going to run it and ensure we don't commit unformatted code.

Balaji028:
Hi team , working on this issue .

Balaji028:
Hi @p0deje  / @sh3pik , Can you Please review this #64  and suggest amendments , so that I can work on it

Balaji028:
Sorry , there is an error coming while running make command trying to fix it , once done will update

Balaji028:
able to sort the issue .

PR #64, opened 2025-04-15
Ci: Added  Makeformat Command in Ci.yml

Balaji028:
#58 , Added `make format` as Run command in Ci.yml file .

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 check-format:
 	poetry run autoflake --check-diff .
 	poetry run black --check --diff .
-	poetry run flake8 alumnium examples
+	poetry run flake8 alumnium examples tests
 	poetry run isort --check .
 	poetry run pyprojectsort --diff
 


.github/workflows/ci.yml
```diff
@@ -61,6 +61,7 @@ jobs:
       - if: matrix.driver == 'playwright'
         run: poetry run playwright install chromium
       - run: Xvfb ${{ env.DISPLAY }} -screen 0 1600x1200x24 &
+      - run: make format #This runs all the Linters as Mentioned in Makefile
```
sh3pik: @Balaji028, `make format` will reformat the code "locally." This means that the CI will not fail, and the files will be silently updated. 

However, we want to ensure we commit only the formatted code. One idea for archiving it is to have a separate make command, e.g., `make check-format`, that would fail on each linter instead of reformatting the code. If possible, you can try providing a custom error message on the failure, like "please run make format on your changes!" or something.
WDYT?

Balaji028: @sh3pik  , Thanks for suggesting  in detail . I will add `make check-format` for each of the linter and provide a custom error message on failure in CI.yml . I will add check-format additional command in existing Makefile in parent directory and in CI  will make use of it using `${{success() }}` / `${{failure() }}` and depending on that we will provide the message which you have suggested . Could you please verify if the approach is fine or if it needs to be changed can you please suggest ?

cgoldberg: The new `Makefile` entry should probably be:
```
check-format:
	poetry run autoflake --check-diff .
	poetry run black --check --diff .
	poetry run isort --check .
	poetry run pyprojectsort --diff
```

This will return a non-zero status if any linter fails, and also output diffs of violations.

sh3pik: @Balaji028 sounds good!

Balaji028: working on that , once done will inform on this .

p0deje: Also, we don't need to run it for all combinations of drivers and models - let's better have a separate job in ci.yml which runs just once.

cgoldberg:
@Balaji028 flake8 was added as a linter today, so remember to include that in the `check-format` target.

sudoCommit:
Hi, you could use Pre-commit hooks, so that all formats are checked before pushing code.

Or Use both!
- Pre-commit hooks: catch and fix before committing.
- CI format check: enforce formatting in pull requests (in case someone didn't install pre-commit locally).

Balaji028:
Thanks @cgoldberg , @sh3pik  , @p0deje  for Suggesting and helping me . These are the changes I have made .

1) Added Check-format in `Makefile `with `autoflake `, `black `, `isort `, `pyprojectsort`, `flake8 `linters .
2) Added new group for Formatters in pyproject.toml file named as `[tools.poetry.group.format.dependencies]` which includes all the above ones .
3) in `CI.yml`  added a new Job named `Format Checker `  , while installing poetry linter dependencies I haven't specified venv as specified in test job .

Please review this and Suggest me where I can do Improvements , so that I can make better .

Makefile
```diff
@@ -10,6 +10,13 @@ test:
 	poetry run behave
 	poetry run pytest
 
+check-format:
+	poetry run autoflake --check-diff .
+	poetry run black --check --diff .
+	poetry run isort --check .
+	poetry run flake8 .
```
sh3pik: Let's keep the same as in `format` command:
```suggestion
	poetry run flake8 alumnium examples
```
Otherwise, it'll validate the virtual environment folder as well

.github/workflows/ci.yml
```diff
@@ -25,7 +25,24 @@ env:
...
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version-file: .tools-versions
+      - run: pip install poetry==$(cat .tool-versions | grep poetry | cut -d" " -f2)
+      - run: poetry install --only format
+      - run: make check-format
+      - if: failure()
+        run: echo "please run make format on your changes!"
```
sh3pik:
```suggestion
        run: echo "Please run 'make format' on your changes!"
```

sh3pik (commented):
@Balaji028, thanks for the changes! Left two comments. Other than that, it looks good to me!

sh3pik:
@sudoCommit, thanks for the idea of pre-commit hooks! I'll add it to the issues list 👍

Makefile
Balaji028: Thanks for suggesting , amended the same

.github/workflows/ci.yml
Balaji028: Thanks for suggesting ! amended the same

Balaji028:
Thanks @sh3pik  for reviewing and suggesting the changes ,  Once can you Please review .

cgoldberg (changes requested):
- The new dependency group in `pyproject.toml` needs to add `flake8-pyproject`.
- To match the other job names in `ci.yml`, `Formatchecker` shouldn't be capitalized (so change it to `formatchecker` or `check-format`).
- The job should have a `name` field.
- To be really pedantic, the `check-format` target should come before `format` in `Makefile` to keep things alphabetical... and the lint jobs within `check-format` should be in alphabetical order to match the `format` target.

cgoldberg (approved):
LGTM

cgoldberg:
@Balaji028 I have a PR open (#69) that edits the `flake8` command in the `Makefile`.

It adds the `tests` directory:

`poetry run flake8 alumnium examples tests`

Can you make sure this is changed in your new `check-format` target as well?

sh3pik (approved):

Balaji028:
Hi @sh3pik  , I see `Code Check Style` CI job  is failing because of unable to find the` . tools-version `  file , I am trying to replicate that issue .

Balaji028:
The issue , is because of extra 's' being appended in `.tool-versions` in `ci.yml`  , because of this CI job is failing . sorry for this , have amended now

PR #68, opened 2025-04-18
Improve logging in alumni.py

LukysYeep:
I added more logging to functions using RichHandler in file alumni.py


p0deje:
Thank you for the attempt, it would be great to have more PRs from you, but for now I'm closing this in favor of #74, which is a more complete implementation.

PR #71, opened 2025-04-21
CI: Added additional Run command poetry lock  and updated pyproject.toml

Balaji028:
Fixes CI failure completely  , after #70  CI failed because of  `poetry.lock` file  not being updated while running the test cases as a result `poetry run --only format` got failed , after a investigation added poetry lock run command before installing dependencies which fixes the `Code check Style job` .

This `Code Check Style Job` has been tested thoroughly and the job result  is https://github.com/Balaji028/alumnium/actions/runs/14579877484/job/40894029134

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,9 +25,8 @@ env:
   DISPLAY: :99
 
 jobs:
-
-  formatchecker:
-    name: Code Style Check
+  check-format:
+    name: Format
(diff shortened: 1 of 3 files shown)


sh3pik:
@Balaji028, thanks for the fixes! 
@p0deje updated some parts in https://github.com/alumnium-hq/alumnium/pull/72. Could you have a look and check if you want to keep the current changes? If so, could you rebase? 🙏

Issue #61, opened 2025-04-12
Introduce per-model prompts


p0deje:
Right now all agents share the same prompts for all supported models. It works fine for now, but inevitably we'll have to introduce model-specific prompts to tune the behavior. For example, if we are to add Gemma 3 support, we'll have to use a custom prompt with explicit function declarations along with LangChain `bind_tools(...)`.

This issue is to refactor current agent prompts to have per-model system/user prompts. We would end up with something like this:

```
alumnium/agents/
├── actor_agent.py
└── actor_prompts
    ├── google
    │   ├── system.md
    │   └── user.md
    └── openai
        ├── system.md
        └── user.md
```

p0deje:
Finalized in https://github.com/alumnium-hq/alumnium/commit/2e79dbaeaf667cedce2cb3582065baae9d80bac4

PR #73, opened 2025-04-22
[Draft] feat(#61)| Introduce per-model prompts

VinayakaMayura:
Implements Task #61: Introduce per-model prompts.

This PR proposes a solution for implementing per-model prompts as discussed in issue #61.

I'd appreciate your feedback on the approach I've taken. Please let me know if you see any alternative solutions or areas for improvement.

I will update the prompts for each model if the approach looks good


p0deje:
Overall, I like the approach. There is a bit of duplication in prompts, but it's expected and we can address this in follow-up PRs. Can you please do the same for all agent classes?

p0deje:
To speed things up a bit, I'm going to merge this and apply the same approach to other agent classes. Thank you for the proposal!

PR #78, opened 2025-05-04
feat: support `mistral-small-31:24b` on ollama

p0deje:
This is the very first fully local model support. Mistral Small 3.1 24B was chosen because it is a great small model that supports both vision and tool calling on Ollama.

In testing on my MacBook with M1 Pro / 32GB  RAM, it passes examples, though it is very slow.

```
$ ALUMNIUM_MODEL=ollama poetry run behave
...
1 feature passed, 0 failed, 0 skipped
8 scenarios passed, 0 failed, 0 skipped
49 steps passed, 0 failed, 0 skipped, 0 undefined
Took 49m49.574s
````

It's not being tested on CI yet because GHA runners cannot run the model. We'll need to add proper testing later. We can potentially test model capabilities by using the Mistral AI platform, but that could be different from the Ollama implementation. Another option is to find a cloud-based solution that can run the model with Ollama and expose an API.


p0deje:
@TrisShare I've seen some issues with structured output used for `check/get`, did it happen to you too? I believe it might be fixed via https://github.com/langchain-ai/langchain/pull/32432/

PR #82, opened 2025-05-11
fix: environment variable setup for logutils

ZahidMKhan:
- adds a fallback default for `ALUMNIUM_LOG_PATH` env variable
- fixes `ALUMNIUM_LOG_LEVEL` env variable configuration


p0deje:
Thank you!

PR #86, opened 2025-05-14
feat: per-test LLM caching

### LINKED ISSUE #59: Implement caching
1. Add API to save cache into a file (sqlite or filesystem cache).
2. This API is supposed to be called in a hook that saves the cache only if the test has passed.

p0deje:
This commit adds LLM caching support in Alumnium implemented on top of SQLite to simplify cache saving/restoring on CI.

The overall idea for the test is the following:
1. Cache is enabled by default for all LLM communications.
2. It automatically starts tracking the requests to LLM and saves them to a database transaction. This transaction effectively serves as an in-memory cache allowing to reuse of the cache during a test execution.
3. Once the test is completed, the user should do one of the following:
  a. _save_ the cache gathered during the test _if the test successfully passes_, or
  b. _discard_ the cache _if the test failed_.
4. Saving the cache would commit the transaction to the database file so subsequent/parallel tests can use it.
5. Discard the cache would roll back the transaction so subsequent/parallel tests would attempt to communicate with LLMs again.

Effectively, a test serves as an oracle. A passing test indicates that Alumnium performed do/check/get correctly. A failing test indicates the opposite - something was wrong during execution so it doesn't make sense to persist LLM choices.

There is still a lot to be done with the whole caching (e.g. cache max size, FIFO or LRU, etc.), so this is just a foundation.

Fixes #59.


alumnium/caching.py
```diff
@@ -0,0 +1,62 @@
+from os import getcwd
```
sh3pik: @p0deje could you rename the file to `cache.py` to be consistent with the class name?

alumnium/caching.py
```diff
@@ -0,0 +1,62 @@
...
+from sqlalchemy.engine.base import Engine
+from sqlalchemy.orm import Session, declarative_base
+
+Base = declarative_base()
+
+
+class Schema(Base):
+    __tablename__ = "alumnium_cache"
+    prompt = Column(String, primary_key=True, nullable=False)
+    llm_string = Column(String, primary_key=True, nullable=False)
```
sh3pik: Will it be a combined primary key (prompt + llm)?

alumnium/caching.py
```diff
@@ -0,0 +1,62 @@
...
+class Cache(BaseCache):
+    def __init__(self, db_path: str = ".alumnium-cache.sqlite", schema: Type[Schema] = Schema):
+        self.engine = create_engine(f"sqlite:///{getcwd()}/{db_path}")
+        self.schema = schema
+        self.schema.metadata.create_all(self.engine)
+
+        self.session = Session(self.engine)
+
+    def save(self) -> None:
+        self.session.commit()
```
sh3pik: Can we wrap it in try-except for the database errors like "primary key is violated"? I'm worried that in parallel execution, we could go into a race condition when two tests execute the same prompt + llm and they both try to commit it into the db. I think we should just ignore such failures and proceed with the tests. WDYT?

alumnium/caching.py
p0deje: No clue! I just did the same thing as Langchain https://python.langchain.com/api_reference/_modules/langchain_community/cache.html#FullLLMCache :)

sh3pik: ChatGPT says it will!

sh3pik (approved):

Issue #93, opened 2025-05-16
Flexible model provided by the provider


sangcnguyen:
Hi, are there options to select other models provided by the provider? In my case, I want to experiment with Gemini 2.0 Flash Lite instead of Gemini 2.0 Flash.

p0deje:
@sangcnguyen There are not at the moment - mainly because we design the system and prompts for specific models. In your particular case, however, I think the same prompts should work. I'll prepare a PR that allows customizing the model via an environment variable `ALUMNIUM_MODEL` and/or `Alumni` class constructor. Are you going to use Google AI Studio or Vertex AI?

sangcnguyen:
I use Google AI Studio. I'd prefer to pass a specific model name using the `ALUMNIUM_MODEL` environment variable.

p0deje:
I will implement it in the following way:

```bash
export ALUMNIUM_MODEL="<provider>/<model>"
```

So in your case it would be:

```bash
export ALUMNIUM_MODEL="google/gemini-2.0-flash-lite"
```

sangcnguyen:
Thank you! It's super useful.

richard-devbot:
@p0deje Are we actually using --vision capabilities - if not we can enhance this - so that models can also see atomic numbers if that's a issue to find out by html

p0deje:
This is implemented in v0.10.0

sangcnguyen:
Thank you for quick update, when will the site update the [configuration](https://alumnium.ai/docs/getting-started/configuration/) accordingly to reflect this change?

p0deje:
I'll push an updated docs probably tomorrow, in the meantime, you can use it as described in https://github.com/alumnium-hq/alumnium/issues/93#issuecomment-2885551263

PR #95, opened 2025-05-18
feat: support custom LLM invocation

Balaji028:
- As per #93 , @p0deje  mentioned in comments , that we are going to support users with flexible models provided in the following format  `export ALUMNIUM_MODEL="<provider>/<model>"` .
- have designed two new methods in models.py file named `Model.provider` returns `enum `and `Model.name` which returns the `string ` value of the provider .
- in `alumni.py` , replaced the existing variable` self.model` with `self.provider`  and created a new variable `self.name` .
- for eg: if user sets `ALUMNIUM_MODEL `as "`google/gemini-2.0-flash-lite`" , in `self.model` `google `value is stored and in `self.provider` `gemini-2.0-flash-lite` is stored and passed to constructor , as a fallback implementation if user sets google only then the default model is being taken from `models.py`.

### DIFF
--- a/alumnium/alumni.py
+++ b/alumnium/alumni.py
@@ -25,7 +25,8 @@ else:
 
 class Alumni:
     def __init__(self, driver: Page | WebDriver, model: Model = None):
-        self.model = model or Model.load()
+        self.provider = model or Model.load()
+        self.name = Model.name()
         if isinstance(driver, WebDriver):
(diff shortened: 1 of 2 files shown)


Balaji028:
@p0deje / @sh3pik , could you Please review this and let me know the further amendments .

p0deje:
Thank you!

PR #102, opened 2025-05-27
feat: support appium-xcuitest

p0deje:
This PR lands an initial implementation of Appium support. The architecture is similar to Selenium/Playwright:

1. Fetch the accessibility tree from the app (`driver.page_source`). 
2. Clean up the tree and convert it to XML.
3. Feed this tree as part of the prompt to LLMs.

There are no changes necessary to agents, though eventually, the list of supported tools needs to change (e.g. we need to add support for `swipe` and remove `hover`). There are a lot of challenges in testing that it all works and I had to check in a built version of https://github.com/ayodejiayankola/To-Do-App-SwiftUI. It's unclear which license it is, so we might need to remove it if it's not a permissive one.

Once this is merged, the `uiautomator2` driver that adds Android support can be implemented in a similar manner. It's going to need its own `UIAutomatorAccessibiltyTree` class, but the rest should _just work_.

Here is a quick demo:

https://github.com/user-attachments/assets/76b44f14-b9c7-40ef-8fad-0f7e840950ee

---------

I've reverted CI changes, but the Appium CI jobs were passing during the work on the PR:


.tool-versions
```diff
@@ -1,2 +1,2 @@
 poetry 2.1.1
-python 3.10.16
+python 3.10.11
```
p0deje, line 2: Downgraded since there is no version for macOS arm64.

alumnium/accessibility/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .accessibility_element import AccessibilityElement
+from .chromium_accessibility_tree import ChromiumAccessibilityTree
+from .xcuitest_accessibility_tree import XCUITestAccessibilityTree
```
p0deje, line 3: Both of these were pretty much vibe-coded, so the result is working but needs heavy refactoring and cleanup.

alumnium/drivers/appium_driver.py
```diff
@@ -0,0 +1,77 @@
...
+
+    @property
+    def aria_tree(self) -> XCUITestAccessibilityTree:
+        return XCUITestAccessibilityTree(self.driver.page_source)
+
+    def click(self, id: int):
+        self._find_element(id).click()
+
+    def drag_and_drop(self, from_id: int, to_id: int):
+        # TODO: Implement drag and drop functionality
```
p0deje, line 28: I'd prefer to land this PR first and then work on implementing other methods with tests.

Issue #113, opened 2025-05-29
Finish Appium implementation


p0deje:
The Appium driver is currently missing implementation for the following actions:

- `drag_and_drop`
- `press_key`
- `select`
- `swipe`

Additionally, we need to change the tools so that they:

- include `hover` only on Selenium/Playwright driver
- include `swipe` on Appium driver

Balaji028:
@p0deje  , Sir i have a doubt in `swipe` API implementation in `appium_driver.py`, in the arguments  it is mentioned only one parameter , but the underlying  Appium API needs four arguments `start_ x` , `start_y` , `end_x,` `end_y` . while implementing it shall i add these parameters sir ? 

https://github.com/alumnium-hq/alumnium/blob/e64e75c58d20944ed5f652016d1399d923d78d06/alumnium/drivers/appium_driver.py#L50

Balaji028:
@p0deje  sir , for implementing the `press_key` functionality can I use this API  https://appium.github.io/python-client-sphinx/webdriver.extensions.html#webdriver.extensions.keyboard.Keyboard.hide_keyboard , under `key_name` parameter can I try to send the enum key value sir

Balaji028:
@p0deje , sir I am bit confused here . `send_keys` takes string argument . can I give some thing like this for keys mentioned in `keys.py` . Eg for `Backspace : driver.send_keys("\b")` . could you please suggest sir ?

p0deje:
@Balaji028 you can pass `Keys` enum to send_keys which would work just fine. I've just tried this on Appium:


```python
from selenium.webdriver.common.keys import Keys

driver.switch_to.active_element.send_keys(Keys.ENTER)
driver.switch_to.active_element.send_keys(Keys.DELETE)
```

PR #116, opened 2025-06-06
feat: appium drag&drop and press_key

Balaji028:
- Added the mentioned API methods to `appium_driver.py` file as per #113 except the `swipe API`.
- for implementing `drag_and_drop` , I used the  existing implementation from `selenium_driver.py` . I tried implementing with `action_helpers` class from appium but not sure whether it is best approach and reverted it .
- similarly for implementing `select` I followed the selenium_driver module.


alumnium/drivers/appium_driver.py
```diff
@@ -25,16 +28,21 @@ def click(self, id: int):
         self._find_element(id).click()
 
     def drag_and_drop(self, from_id: int, to_id: int):
-        # TODO: Implement drag and drop functionality
-        pass
-
-    def hover(self, id: int):
-        # TODO: Remove hover tool, it's not supported in Appium
-        pass
+        actions = ActionChains(self.driver)
```
p0deje: You mentioned in https://github.com/alumnium-hq/alumnium/issues/113#issuecomment-2944128461 that Appium already implements drag_and_drop, so maybe we should just use it instead?

alumnium/drivers/appium_driver.py
```diff
@@ -25,16 +28,21 @@ def click(self, id: int):
...
-        # TODO: Implement press key functionality
-        pass
+        if key == Keys.BACKSPACE:
+            self.driver.switch_to.active_element.send_keys(Keys.BACKSPACE)
+        elif key == Keys.ENTER:
+            self.driver.switch_to.active_element.send_keys(Keys.ENTER)
+        elif key == Keys.ESCAPE:
+            self.driver.switch_to.active_element.send_keys(Keys.ESCAPE)
+        elif key == Keys.TAB:
+            self.driver.switch_to.active_element.send_keys(Keys.TAB)
```
p0deje: This won't work because `key` is https://github.com/alumnium-hq/alumnium/blob/f845c08c5bedb8c54bcb4325067c69a8592f3fc9/alumnium/drivers/keys.py.

I think we can use the same approach as in Selenium driver:

```suggestion
        keys = []
        if key == Key.BACKSPACE:
            keys.append(Keys.BACKSPACE)
        elif key == Key.ENTER:
            keys.append(Keys.ENTER)
        elif key == Key.ESCAPE:
            keys.append(Keys.ESCAPE)
        elif key == Key.TAB:
            keys.append(Keys.TAB)

        self.driver.switch_to.active_element.send_keys(*keys).perform()
```

alumnium/drivers/appium_driver.py
```diff
@@ -44,8 +52,11 @@ def screenshot(self) -> str:
         return self.driver.get_screenshot_as_base64()
 
     def select(self, id: int, option: str):
-        # TODO: Implement select functionality
-        pass
+        element = self._find_element(id)
+        # Anthropic chooses to select using option ID, not select ID
+        if element.tag_name == "option":
+            element = element.find_element(By.XPATH, ".//parent::select")
```
p0deje: That won't work on Appium because there are no `select` elements. Let's remove Anthropic condition.

alumnium/tools/hover_tool.py
```diff
@@ -1,4 +1,6 @@
 from pydantic import BaseModel, Field
```
p0deje: Please revert these changes. We actually need to skip `hover` from `alumnium.tools.ALL_TOOLS` on Appium, but we can do this in follow-up PRs.

alumnium/drivers/appium_driver.py
Balaji028: ok sir , I will use the `drag_and_drop` from Appium

alumnium/drivers/appium_driver.py
Balaji028: @p0deje  sir , I saw a discussion in Appium for selecting a specific element in `dropdown` https://discuss.appium.io/t/how-we-can-use-best-approach-to-automate-dropdown-selection-in-flutter-app-using-appium/45801/2 , in that it mentioned that we need to use `XCUIElementTypePickerWheel` as element type . I tried to some thing like this sir 

``` python
 def  select(self, id:int, option:str):
      drop_down_id = self._find_element(id)
      if drop_down_id .tag_name == "XCUIElementTypePickerWheel":
           drop_down_id .send_keys(option)
      else:
         raise ValueError("the dropdown is not clickable")
 ```
can you please review this  approach sir , I guess I need to work on `date-pickers`, `static `, `dynamic `dropdowns for select API.

alumnium/drivers/appium_driver.py
Balaji028: @p0deje  sir , I will stick with `selenium_driver.py` approach .

alumnium/drivers/appium_driver.py
p0deje: @Balaji028 It's hard to say since there are no tests. Let's revert these changes for now so we could merge the rest.

Balaji028: ok sir , I will remove the select functionality totally

Balaji028:
@p0deje  sir , I have added two methods `drag_and_drop` and `press_key` . i have reverted the `select` method.

p0deje (approved):

p0deje:
Thank you!

PR #122, opened 2025-06-26
Added Appium example to conftest.py and built out AppiumDriver's url property

shbenzer:
## Summary
- Adds an appium driver example to the pytest examples' conftest.py (most recent ios)
- Adds url property  functionality for AppiumDriver

## Motivation
Useful for appium development

## Related Issue(s)
Continues #113 

## Type of Change
Mark the relevant options.

- [ ] Bug fix  
- [x] New feature (non-breaking changes, test coverage, refactoring)
- [ ] Breaking change (fix, refactoring, or feature that would cause existing functionality to change)
- [ ] Cleanup (documentation, formatting)

## Code or UI Demos (if applicable)
To run the appium server for testing

```bash
appium --address 127.0.0.1 --port 4724 --base-path /wd/hub --log-level info
```

passed calculator local test:

```bash
============================= test session starts ==============================
platform darwin -- Python 3.11.5, pytest-8.3.4, pluggy-1.5.0
rootdir: /Users/simonbenzer/Desktop/Repos/Github/alumnium
configfile: pyproject.toml
plugins: html-4.1.1, metadata-3.1.1, retry-1.7.0, anyio-4.8.0
collected 1 item                                                               

examples/pytest/calculator_test.py .                                     [100%]

============================== 1 passed in 54.05s ==============================
(appium-alumnium) (base) simonbenzer@Simons-MacBook-Pro alumnium % 
```

## Checklist
Please verify the following before submitting:

- [x] I have read the [Contributing Guidelines](https://github.com/alumnium-hq/alumnium/blob/main/CONTRIBUTING.md)
- [ ] I have added or updated relevant documentation
- [x] I have written or updated necessary tests
- [x] I have tested the changes locally
- [x] The changes follow the project's coding style and structure
- [x] I have not included any sensitive or credential-related information
- [x] I have ensured these changes maintain or improve accessibility (for UI changes)
- [x] I have considered security implications of these changes

## Additional Notes
- This is really cool @p0deje , awesome work
- Was originally going to see if press_key needed further development, but had some fun getting sidetracked

PR #123, opened 2025-06-28
feat: appium title extraction

p0deje:
Follow-up to #122

### DIFF
--- a/examples/pytest/conftest.py
+++ b/examples/pytest/conftest.py
@@ -22,7 +22,6 @@ def driver():
         yield driver
     elif driver == "appium":
         options = XCUITestOptions()
-        # https://github.com/ayodejiayankola/To-Do-App-SwiftUI
         options.automation_name = "XCUITest"
         options.bundle_id = "com.apple.mobilesafari"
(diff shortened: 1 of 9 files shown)


examples/pytest/conftest.py
```diff
@@ -22,19 +21,15 @@ def driver():
...
-        )
-        driver = AppiumWebdriver.Remote(command_executor="http://127.0.0.1:4724/wd/hub", options=options)
+        options = XCUITestOptions()
+        options.automation_name = "XCUITest"
+        options.bundle_id = "com.apple.mobilesafari"
+        options.device_name = "iPhone 15"
+        options.new_command_timeout = 300
+        options.no_reset = True
+        options.platform_name = "iOS"
+        options.platform_version = "17.4"
```
sh3pik, line 31: JIC can we use a higher iOS version?

sh3pik (approved):

examples/pytest/conftest.py
p0deje: It's not available at GHA yet.

Issue #112, opened 2025-05-29
Support Android with Appium and UIAutomator2


p0deje:
Since #102 we have a basic implementation of Appium with the XCUITest driver on iOS platforms. The next step is to implement similar support for the UIAutomator2 driver so Alumnium can be used with Appium on Android.

The implementation is likely going to be similar to XCUITest because Appium provides consistent APIs for actions, but we'll need a custom accessibility tree parser similar to `XCUITestAccessibilityTree`.

Balaji028:
Hi team , I have started analyzing the `XCUITestAccessbilityTree` module currently developed for ios . I will start designing the same for Android ,once I get full understanding of the accessibility tree module , might take some time but I will try to implement it .

p0deje:
@Balaji028 Not sure, I've registered at BrowserStack and followed the steps in https://github.com/browserstack/python-appium-app-browserstack/blob/master/README.md - everything worked fine. This is the final script I have:

```python
from appium import webdriver
from appium.options.android import UiAutomator2Options
from appium.webdriver.common.appiumby import AppiumBy
from selenium.webdriver.support.ui import WebDriverWait
from selenium.webdriver.support import expected_conditions as EC

options = UiAutomator2Options().load_capabilities(
    {
        # Specify device and os_version for testing
        "platformName": "android",
        "platformVersion": "9.0",
        "deviceName": "Google Pixel 3",
        # Add your caps here
    }
)

driver = webdriver.Remote("http://127.0.0.1:4723/wd/hub", options=options)
search_element = WebDriverWait(driver, 30).until(
    EC.element_to_be_clickable((AppiumBy.ACCESSIBILITY_ID, "Search Wikipedia"))
)
print(driver.page_source)  # Print the page source for debugging
driver.quit()
```

And this is how I run it:

```bash
browserstack-sdk python browserstack_sample.py
```

-------------


Running the steps above, I managed to get the accessibility tree - https://gist.github.com/p0deje/782f47d3f19f66b7dd719df99f39b09d. As I am looking through it, it seems to have all the necessary information (text, view names, displayed property, etc)

p0deje:
@Balaji028 I think all we need is a tag/name/label/value to build an XPath, you can ignore the rest

PR #131, opened 2025-07-19
feat:  added support for android_ui_automator2 

Balaji028:
- Implements #112 
- In `ui_automator2_accessbility_tree.py` file added functionality to make sure  the output of `driver.pagesource` is valid so that it can be parsed (some times when the page source is huge there are multiple xml declarations present in string , identified them and removed using regular expression and wrapped up all the elements under root tag . For example `<root> <hierarchy> </root>`) .
- In `parse_element` method took all the attributes from xml string and  added them to properties variable for Node for each id .
- for `find_element_by_id` , for Android since the attributes vary from IOS , in `accessbility_element.py` added new values like `androidresourceid` , `androidclass`, `androidtext` , `androidcontentdesc` , `androidbounds`.  and used them in `appium_driver.py` while constructing xpath .
- In `alumni.py` added mobile_platform string default to `Android` .
-  added tests  for `element_by_id` .


alumnium/alumni.py
```diff
@@ -23,11 +23,15 @@
 
 
 class Alumni:
-    def __init__(self, driver: Page | WebDriver, model: Model = None):
+    def __init__(self, driver: Page | WebDriver, model: Model = None, mobile_platform: str = "Android"):
```
p0deje: I don't think we need it. We should instead attempt to check what Appium driver is passed via `driver` (iOS or Android). I believe there are some properties we can check that are available in `driver` variable...

alumnium/drivers/appium_driver.py
```diff
@@ -10,6 +10,7 @@
 
 
 from alumnium.accessibility import XCUITestAccessibilityTree
+from alumnium.accessibility.uiautomator2_accessibility_tree import UIAutomator2AccessibiltyTree
```
p0deje: You can add `UIAutomator2AccessibiltyTree` to `__init__.py` so it can be imported from `alumnium.accessibility` directly.

alumnium/drivers/appium_driver.py
```diff
@@ -19,12 +20,16 @@
 
 
 class AppiumDriver(BaseDriver):
-    def __init__(self, driver: Remote):
+    def __init__(self, driver: Remote, mobile_platfrom: str):
```
p0deje: Likewise, we should determine the platform using `driver` variable instead of an explicit flag.

tests/accessibility/test_uiautomator2_accessibility_tree.py
```diff
@@ -0,0 +1,22 @@
...
+
+def tree(filename: str) -> UIAutomator2AccessibiltyTree:
+    with open(Path(__file__).parent.parent / "fixtures" / f"{filename}.xml", "r") as f:
+        xml = f.read()
+    return UIAutomator2AccessibiltyTree(xml)
+
+
+@fixture
+def simple_tree() -> UIAutomator2AccessibiltyTree:
+    return tree("uiautomator2_accessibility_tree")
```
p0deje, line 17: Can you please add a test that shows asserts the output of `to_xml()` method?

p0deje (commented):
I think it's a great start, let's proceed with adding `to_xml` so we can actually compare with what we have on iOS.

tests/accessibility/test_uiautomator2_accessibility_tree.py
Balaji028: sir , i tried locally , but the test is failing for to `to_xml()` method trying to understand it why it is failing

Balaji028: @p0deje  , Sir I am getting stucked while running test for `to_xml()` function . Every time when i run it it is getting failed while comparing it with xml string provided in test . I used `strip()` function to remove new lines and unwanted spaces but still it failing . Could you please help me sir how to solve this issue . 

image

output from to xml sir , https://gist.github.com/Balaji028/99a00db55d8cabf9cfeaacc513eac82b  when I assert this with string value test is failing .

p0deje: @Balaji028 Can you run pytest with `-vv` and paste the output to gist?

Balaji028: @p0deje  , Sir i tried running it  https://gist.github.com/Balaji028/327d9af622d87dcd99ca919715efe0ed

Balaji028: @p0deje Sir , I have made amendments to the `to_xml()`  function , now when i run test it is failing , because of some UTF-8 text characters not getting decoded properly some unnecessary characters being interpreted when comparing it , tried to run this in colab where it is able to decode properly while comparing two xml strings but in VS code terminal i am getting differences like this . could you please suggest sir how to overcome this issue? , rest everything is working fine except this.

```
-                                        <TextView id="35" text="The Vera C. Rubin Observatory in Chile releases 
the first light images from its new 8.4-meter (28 ft) telescope." resource-id="org.wikipedia.alpha:id/horizontal_scroll_list_item_text" />
+                                        <TextView id="35" text="The Vera C. Rubin Observatory in Chile releases 
the first light images from its new 8.4-meter (28Â ft) telescope." resource-id="org.wikipedia.alpha:id/horizontal_scroll_list_item_text" />

here after (28 ft) this got added (28Â ft) got added .
```
image

Balaji028: @p0deje  sir , i have added test for `to_xml()` it has passed now , now in the xml format , I have  considered attribute `resourceid` along with node id for each xml element to make xml look better . I tried adding text content as well for `textview` type elements but test failed locally in my machine because of encoding and decoding issue while comparing to xml strings especially for unicode characters , to overcome this issue i tried using `lxml` ,`unicodedata` module but still because of space character issue in text attribute , I am not sure how to solve this removed text content attribute because of this issue , based on review i will progress on this sir .

image

tests/accessibility/test_uiautomator2_accessibility_tree.py
```diff
@@ -0,0 +1,270 @@
...
+                                  <TextView id="93" />
+                                  <ImageView id="94" resource-id="org.wikipedia.alpha:id/voice_search_button" />
+                                </LinearLayout>
+                              </FrameLayout>
+                              <FrameLayout id="95">
+                                <RelativeLayout id="96">
+                                  <FrameLayout id="97" resource-id="org.wikipedia.alpha:id/view_list_card_header">
+                                    <LinearLayout id="98">
+                                      <LinearLayout id="99">
+                                        <ImageView id="100" resource-id="org.wikipedia.alpha:id/view_card_header_image" />
```
p0deje, line 171: I think it would be great to include relevant attributes such as text and content-desc if they are not empty strings, and add clickable if it's true. Without these attributes, I feel like LLM wouldn't be able to understand which elements to interact with.

Balaji028:
@p0deje  sir , I have added the attributes `text` , `content-description` , ` clickable` while constructing tree form Node object . In test file correspondingly I have added `proper_format` helper function which ensure proper encoding and decoding of strings to UTF-8 format .I tried to run test locally it passed sir .

p0deje:
@Balaji028 There are some failures on CI related to UTF-8 encoding - https://github.com/alumnium-hq/alumnium/actions/runs/16695964450/job/47260134723?pr=131. Frankly, I don't think you need to touch encoding at all - I personally didn't do that in other tests.

Balaji028:
@p0deje  sir , sorry I have now  removed all the unnecessary encoding and decoding totally and  have made it clean now . The two tests passed locally . Some times while comparing `xml_output` with `expected_output` there is some issue with Unicode characters in text content attribute especially with the space character getting misinterpreted  , in order to overcome this issue , I have used `NFKC` normalization form from `unicodedata` module in tests before passing to `ui_automator2` .  Sorry sir ,previously I had unnecessarily dig deeper and added encoding and decoding sorry for that . 

image

Balaji028:
Ok sir , working on that

Issue #105, opened 2025-05-27
Introduce client-server architecture


p0deje:
We want to support multiple languages (Ruby, JavaScript) and we need to centralize the LLM communication within a single place so we don't re-implement all the prompts/techniques/caching/etc in all languages. This can be done by splitting the architecture into two pieces:

1. The client which is responsible for talking to the browser. It would send a request to the server to `do/check` and receive a response with instructions on what needs to be done in the browser, then execute them.
2. The server which is responsible for talking to LLM. It would accept requests to `do/check` something with the necessary context (ARIA, URL, etc), then communicate to LLM and respond with the instructions on what needs to be done on the client.

A rough API would look like this:

| Request | Response | Description |
|--------|----------|------------|
| POST /sessions | {"sessionId": ...} | Starts a new session on the server | 
| DELETE /sessions/{sessionId} | | Stops the session on the server |
| POST /sessions/{sessionId}/actions <br> {"goal": ..., "aria": ...}  | {"actions": [{"type": "click", "args": {"id": 1}}]} | Plan how to do certain actions on the page  | 
| POST /sessions/{sessionId}/verifications <br> {"statement": ..., "screenshot": ...} | {"result": true, "explanation": "..."} | Perform the verification on the page  | 

In the initial implementation for the Python library, this can be the same package as the "in-memory" server. This means that we are going to serialize the payloads, call endpoints directly as Python methods, and deserialize the response back.

For other languages, the server endpoint will be accessible via an HTTP server that is run as a Docker container (e.g. `docker run -p 8081:8081 alumnium`). We don't have to implement it yet, let's first focus on the splitting and Python library.

PR #133, opened 2025-07-26
feat: separate LLM agents and web/mobile drivers

sh3pik:
It's the first step in having a client-server architecture.
Refactored the code to make agents work with the provided data instead of calling the driver's methods.

Part of issue https://github.com/alumnium-hq/alumnium/issues/105

### DIFF
--- a/alumnium/agents/area_agent.py
+++ b/alumnium/agents/area_agent.py
@@ -1,6 +1,7 @@
 from langchain_core.language_models import BaseChatModel
 from pydantic import BaseModel, Field
 
+from alumnium.accessibility.base_accessibility_tree import BaseAccessibilityTree
 from alumnium.drivers import BaseDriver
 from alumnium.logutils import get_logger
 
@@ -20,16 +21,13 @@ class Area(BaseModel):
 
 
 class AreaAgent(BaseAgent):
-    def __init__(self, driver: BaseDriver, llm: BaseChatModel):
+    def __init__(self, llm: BaseChatModel):
         super().__init__()
-        self.driver = driver
         self.chain = self._with_retry(llm.with_structured_output(Area, include_raw=True))
 
-    def invoke(self, description: str) -> Area:
+    def invoke(self, description: str, accessibility_tree: BaseAccessibilityTree) -> Area:
         logger.info("Starting area detection:")
         logger.info(f"  -> Description: {description}")
-
-        accessibility_tree = self.driver.accessibility_tree
         logger.debug(f"  -> Accessibility tree: {accessibility_tree.to_xml()}")
 
         message = self.chain.invoke(
@@ -38,7 +36,8 @@ class AreaAgent(BaseAgent):
                 (
                     "human",
                     self.prompts["user"].format(
-                        accessibility_tree=accessibility_tree.to_xml(), description=description
+                        accessibility_tree=accessibility_tree.to_xml(),
+                        description=description,
                     ),
                 ),
             ]
@@ -58,8 +57,4 @@ class AreaAgent(BaseAgent):
         logger.info(f"  <- Usage: {message['raw'].usage_metadata}")
         self._update_usage(message["raw"].usage_metadata)
 
-        return {
-            "id": response.id,
-            "explanation": response.explanation,
-            "accessibility_tree": self.driver.accessibility_tree.get_area(response.id),
-        }
+        return {"id": response.id, "explanation": response.explanation}
--- a/alumnium/agents/area_agent.py
+++ b/alumnium/agents/area_agent.py
@@ -1,8 +1,6 @@
 from langchain_core.language_models import BaseChatModel
 from pydantic import BaseModel, Field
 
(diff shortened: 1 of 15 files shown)


alumnium/area.py
```diff
@@ -48,10 +58,12 @@ def check(self, statement: str, vision: bool = False) -> str:
...
         """
-        result = self.retrieval_agent.invoke(
+        result = self.retriever_agent.invoke(
             f"Is the following true or false - {statement}",
-            vision,
-            self.accessibility_tree,
+            self.accessibility_tree.to_xml(),
+            title=self.driver.title,
+            url=self.driver.url,
+            screenshot=self.driver.screenshot if vision else None,
```
sh3pik, line 66: @p0deje The `retriever_agent` takes the screenshot of the whole page. I didn't refactor it here, but we can take a screenshot of the area. AFAIU, it's easy for web, but not mobile.

alumnium/agents/area_agent.py
```diff
@@ -20,16 +21,13 @@ class Area(BaseModel):
...
 
 class AreaAgent(BaseAgent):
-    def __init__(self, driver: BaseDriver, llm: BaseChatModel):
+    def __init__(self, llm: BaseChatModel):
         super().__init__()
-        self.driver = driver
         self.chain = self._with_retry(llm.with_structured_output(Area, include_raw=True))
 
-    def invoke(self, description: str) -> Area:
+    def invoke(self, description: str, accessibility_tree: BaseAccessibilityTree) -> Area:
```
p0deje: Shouldn't accessibility_tree be a string here like in ActorAgent?

alumnium/alumni.py
```diff
@@ -87,9 +93,16 @@ def do(self, goal: str):
...
             goal: The goal to be achieved.
         """
-        steps = self.planner_agent.invoke(goal)
-        for step in steps:
-            self.actor_agent.invoke(goal, step)
+        initial_accessibility_tree = self.driver.accessibility_tree
+        steps = self.planner_agent.invoke(goal, initial_accessibility_tree.to_xml())
+        for idx, step in enumerate(steps):
+            # If the step is the first step, use the initial accessibility tree.
+            accessibility_tree = initial_accessibility_tree if idx == 0 else self.driver.accessibility_tree
```
p0deje, line 98: I don't get this condition, I think we used to re-fetch tree on each step. We can change this approach actually, I've thought about passing the same tree to each actor agent call. If you think it makes sense, let's do in a separate PR.

alumnium/area.py
p0deje: @sh3pik Good catch! Let's create a ticket so we don't forget.

alumnium/area.py
```diff
@@ -67,4 +79,28 @@ def get(self, data: str, vision: bool = False) -> Data:
...
-        return self.retrieval_agent.invoke(data, vision, self.accessibility_tree).value
+        return self.retriever_agent.invoke(
+            data,
+            self.accessibility_tree.to_xml(),
+            title=self.driver.title,
+            url=self.driver.url,
+            screenshot=self.driver.screenshot if vision else None,
+        ).value
+
+    def _execute_tool_call(
```
p0deje: Do you think it makes sense to move this to the tool class itself to avoid duplication with alumni.py?

examples/pytest/table_test.py
```diff
@@ -46,7 +46,7 @@ def test_table_sorting(al, navigate):
     assert table2.get("first names") == ["John", "Frank", "Jason", "Tim"]
     assert table2.get("last names") == ["Smith", "Bach", "Doe", "Conway"]
 
-    table1.do("sort by last name")
+    table1.do("sort by last name once")
```
p0deje: If the issue is that it double clicks, you can add Provider.GOOGLE to `learn` fixture that fixes the same problem on Haiku.

p0deje (approved):
Amazing work, thank you!

alumnium/alumni.py
sh3pik: > I've thought about passing the same tree to each actor agent call
@p0deje, I don't really agree with it. We have `area` now, so you can achieve the save with it. 

What I did here was to avoid taking an unnecessary accessibility tree when executing the very first step. The `planner_agent` doesn't interact with the page/screen, so the `accessibility_tree` remains the same on the first step. Though it may change after it.

examples/pytest/table_test.py
sh3pik: ah, right, I didn't think about it

alumnium/area.py
sh3pik: good point

alumnium/area.py
sh3pik: https://github.com/alumnium-hq/alumnium/issues/136

alumnium/agents/area_agent.py
sh3pik: yes, good catch, missed it

p0deje (approved):

PR #142, opened 2025-07-31
feat!: structured planner in openai

p0deje:
This PR migrates OpenAI models to use a structured chain-of-thought planner similar to the one Llama 4 uses now. It's much more accurate and results in fewer hallucinations, such as attempting to do something not possible in the accessibility tree. As a side effect, the tests that used to exploit this behavior now fail. Overall, it requires the tests to be written more concretely and puts less weight on examples given with `al.learn`.

Some examples of how tests might need more explicit instructions:
- 3964268c4a02d8f1475a1dc5bbd246d54135b844
- c48e41ac67959f9e602d37101b352e7b8fd7cd83

### DIFF
--- a/alumnium/agents/planner_prompts/openai/system.md
+++ b/alumnium/agents/planner_prompts/openai/system.md
@@ -17,7 +17,6 @@ When formulating your actions:
 6. Ground the actions in the accessibility tree provided.
 7. Action "drag and drop" is always performed as a single step.
 8. Always aim to minimize the number of actions. If a single step suffices to accomplish the task, do not break it down further.
-9. When you are given examples, use their actions with minimal modifications based on the accessibility tree.
 
 If you cannot find a way to achieve the goal based on the given accessibility tree, respond with an empty list of actions.
 
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ driver = Chrome()
 driver.get("https://duckduckgo.com")
 
 al = Alumni(driver)
-al.do("search for selenium")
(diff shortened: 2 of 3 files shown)


alumnium/agents/planner_prompts/openai/system.md
```diff
@@ -1,7 +1,32 @@
-You are a helpful assistant that plans what actions should be performed to achieve a task on a webpage based on the accessibility tree of the page given as XML.
-If you don't see a way to achieve the goal on the webpage, reply NOOP. Otherwise, reply with a list of steps separated {separator}. Don't include anything else beyond steps.
-Do not include element id in the step.
-Include element tag name in the step.
-Include element text content if it's present.
-Wrap step arguments except tag name in quotes.
-Each step can start with one of the following: click, drag and drop, hover, press key, select, type.
+You are an AI assistant tasked with planning actions to achieve a specific goal on a webpage based on the accessibility tree provided. The accessibility tree is given as XML and represents the structure and elements of the webpage.
```
sh3pik, line 1: I think `openai` prompts are used as default ones when model/provider doesn't have one. It might be the reason why `mistralai` tests are failing on PR CI. Shall we add a prompt for `mistralai`?

p0deje: Yeah, I will work on enabling a structured planner for Mistral later and will add a separate prompt. It's failing in master anyway - https://github.com/alumnium-hq/alumnium/actions/runs/16767063184/job/47474205485.

PR #143, opened 2025-08-06
feat!: reuse accessibility tree in `do`

p0deje:
This PR adds multiple optimizations that overall increase Appium tests execution up to 3x:

1. Avoids re-fetching the accessibility tree multiple times within `do()` calls. Each fetch can take a few seconds on large React Native apps. Though we generally want to avoid operating on a changing tree, since the planner steps could become incorrect at this point.
2. Migrates XPath locators to iOS Predicates. Those are faster according to https://appium.github.io/appium-xcuitest-driver/latest/reference/locator-strategies/.


p0deje:
Closing since most of the work was pulled in #179 and reusing accessibility tree can be imlpemented in a follow-up.

PR #147, opened 2025-08-11
feat: initial client-server implementation

sh3pik:
## Summary
This PR implements the initial implementation of the monorepo and client-server setup. Current implementation talks to the server directly and doesn't use the HTTP layer.

## Motivation
To support multiple languages(e.g. JS) in the future, the LLM interactions should be moved to the server, so there will be only one implementation of Alumnium prompts, agents and supported models.

## Related Issue(s)
Part of #105

## Type of Change
Mark the relevant options.

- [ ] Bug fix  
- [x] New feature (non-breaking changes, test coverage, refactoring)
- [ ] Breaking change (fix, refactoring, or feature that would cause existing functionality to change)
- [ ] Cleanup (documentation, formatting)

## Evidence

| Description | Evidence |
| --- | --- |
| Package build + install | Screenshot 2025-08-30 at 9 54 39 PM |
| Import + command execution | Screenshot 2025-08-30 at 9 51 16 PM |

### DIFF
--- a/alumnium/server/main.py
+++ b/alumnium/server/main.py
@@ -1,4 +1,5 @@
 import base64
+import importlib.metadata
 from contextlib import asynccontextmanager
 from typing import List
 
@@ -42,7 +43,10 @@ async def lifespan(app: FastAPI):
 
 # FastAPI app
 app = FastAPI(
-    title="Alumnium Server", description="AI-powered test automation server", version="0.13.0", lifespan=lifespan
+    title="Alumnium Server",
+    description="AI-powered test automation server",
+    version=importlib.metadata.version("alumnium"),
+    lifespan=lifespan,
 )
 
 # Add CORS middleware
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,7 @@ packages = [
 ]
 readme = "README.md"
 repository = "https://github.com/alumnium-hq/alumnium"
-version = "0.13.0"  # Also update in server/main.py
+version = "0.13.0"
 
 [tool.poetry.dependencies]
 appium-python-client = "^5.1.1"


alumnium/server/main.py
```diff
@@ -0,0 +1,207 @@
...
+        # Use retriever agent to verify the statement
+        result = session.retriever_agent.invoke(
+            request.statement,
+            request.accessibility_tree,
+            title=request.title,
+            url=request.url,
+            screenshot=screenshot_bytes,
+        )
+
+        return VerificationResponse(result=bool(result.value), explanation=result.explanation)
```
sh3pik: check why `bool(result.value)`

alumnium/server/main.py
```diff
@@ -0,0 +1,207 @@
...
+        return StepResponse(actions=actions)
+
+    except Exception as e:
+        logger.error(f"Failed to execute actions for session {session_id}: {e}")
+        raise HTTPException(
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR, detail=f"Failed to execute actions: {str(e)}"
+        )
+
+
+@app.post("/sessions/{session_id}/verifications", response_model=VerificationResponse)
```
sh3pik:
```suggestion
@app.post("/sessions/{session_id}/retrieve", response_model=RetrieveResponse)
```

alumnium/server/api_models.py
```diff
@@ -0,0 +1,61 @@
...
+    explanation: str
+
+
+class AreaRequest(BaseModel):
+    description: str
+    accessibility_tree: str
+
+
+class AreaResponse(BaseModel):
+    area: str
```
sh3pik:
```suggestion
    area: 
      id: int
      explanation: str
```

alumnium/server/poetry.lock
p0deje, line 1: Do we need this file?

alumnium/server/main.py
```diff
@@ -0,0 +1,246 @@
...
+    logger.error(f"Unhandled exception: {exc}")
+    error_response = ErrorResponse(error="Internal server error", detail=str(exc))
+    return JSONResponse(status_code=500, content=error_response.model_dump())
+
+
+def main():
+    """Main entry point for running the server."""
+    import uvicorn
+
+    uvicorn.run("alumnium.server.main:app", host="0.0.0.0", port=8000, reload=True, log_level="info")
```
p0deje: let's use 8013 (13 is Al element)

alumnium/server/agents/actor_agent.py
```diff
@@ -1,17 +1,17 @@
+from typing import Dict, Type
```
p0deje, line 1: Is this needed? In VS Code, it worked fine with previous code.

alumnium/server/cache/sqlite_cache.py
```diff
@@ -50,7 +50,8 @@ class CacheEntry(Base):
 
 class SQLiteCache(BaseCache):
     def __init__(self, db_path: str = ".alumnium-cache.sqlite"):
-        self.engine = create_engine(f"sqlite:///{getcwd()}/{db_path}", connect_args={"timeout": 15})
+        project_root = Path(__file__).parent.parent.parent
```
p0deje: A cache file should be placed in the current working directory. Is there an issue with the previous code?

alumnium/server/main.py
```diff
@@ -0,0 +1,246 @@
...
+async def lifespan(app: FastAPI):
+    """Application lifespan manager."""
+    logger.info("Starting Alumnium Server")
+    yield
+    logger.info("Shutting down Alumnium Server")
+
+
+# FastAPI app
+app = FastAPI(
+    title="Alumnium Server", description="AI-powered test automation server", version="0.1.0", lifespan=lifespan
```
p0deje: Let's use the same version as in `pyproject.toml`.

alumnium/server/poetry.lock
p0deje: Not sure I get your reply :)

Makefile
```diff
@@ -40,3 +40,24 @@ test-openai:
...
+
+# Installation commands
+install-client:
+	poetry install --with dev
+
+install-server:
+	poetry install --with server,dev
+
+# Server commands
+server-serve:
```
p0deje: nit: serve server sounds odd?


```suggestion
start-server:
```

Makefile
```diff
@@ -40,3 +40,24 @@ test-openai:
...
+	poetry run alumnium-server
+
+server-serve-dev:
```
p0deje:
```suggestion
start-server-dev:
```

Makefile
```diff
@@ -40,3 +40,24 @@ test-openai:
...
+	poetry run python -m alumnium.server.main
+
+# Server tests
+server-test:
```
p0deje:
```suggestion
test-server:
```

alumnium/server/poetry.lock
sh3pik: @p0deje you removed it, right? I agree that it's not needed!

alumnium/server/agents/actor_agent.py
sh3pik: The PR went through a couple of rebases and iterations of changes, so if it works with the previous version, can remove!

alumnium/server/cache/sqlite_cache.py
sh3pik: Yeah, it was putting cache in the /server directory. I think that happened when I started the server. However, now that I'm thinking about it, I believe it should be the correct behaviour. Need to revert the change, good catch!

alumnium/server/agents/actor_agent.py
p0deje: I think we can keep it as-is and maybe go over all typing support in a separate PR.

PR #153, opened 2025-08-12
feat(tool): add "Go Back"

ioachim-hub:
## Summary
This PR introduces a new tool into the framework which is helping user navigate back during testing.

## Motivation
I introduced this update because I needed it on my personal project and I though it might be useful for other people too.

## Related Issue(s)
No related issues.

## Type of Change
Mark the relevant options.

- [ ] Bug fix  
- [X] New feature (non-breaking changes, test coverage, refactoring)
- [ ] Breaking change (fix, refactoring, or feature that would cause existing functionality to change)
- [ ] Cleanup (documentation, formatting)

## Code or UI Demos (if applicable)
I added another test that if validating this.

## Checklist
Please verify the following before submitting:

- [X] I have read the [Contributing Guidelines](https://github.com/alumnium-hq/alumnium/blob/main/CONTRIBUTING.md)
- [ ] I have added or updated relevant documentation
- [X] I have written or updated necessary tests
- [X] I have tested the changes locally
- [X] The changes follow the project's coding style and structure
- [X] I have not included any sensitive or credential-related information
- [ ] I have ensured these changes maintain or improve accessibility (for UI changes)
- [X] I have considered security implications of these changes

### DIFF
--- a/alumnium/drivers/playwright_driver.py
+++ b/alumnium/drivers/playwright_driver.py
@@ -60,6 +60,9 @@ class PlaywrightDriver(BaseDriver):
     def quit(self):
         self.page.close()
 
+    def back(self):
+        self.page.go_back()
+
     @property
     def screenshot(self) -> str:
         return b64encode(self.page.screenshot()).decode()
--- a/alumnium/drivers/selenium_driver.py
+++ b/alumnium/drivers/selenium_driver.py
@@ -62,6 +62,9 @@ class SeleniumDriver(BaseDriver):
     def quit(self):
         self.driver.quit()
 
+    def back(self):
+        self.driver.back()
+
     @property
     def screenshot(self) -> str:
         return self.driver.get_screenshot_as_base64()
--- a/examples/behave/features/environment.py
+++ b/examples/behave/features/environment.py
@@ -82,6 +82,7 @@ def alumnium(context):
                 'click button "x" near the "this is Al" task',
             ],
         )
+
     yield context.al
     context.al.quit()
 
--- a/examples/pytest/conftest.py
+++ b/examples/pytest/conftest.py
@@ -55,6 +55,7 @@ def al(driver):
     al = Alumni(driver)
     if driver_type == "appium":
         al.driver.delay = 0.1
+
     yield al
     al.quit()
 
--- a/alumnium/agents/planner_prompts/deepseek/system.md
+++ b/alumnium/agents/planner_prompts/deepseek/system.md
@@ -9,7 +9,7 @@ Your goal is to determine a series of actions that will accomplish the task desc
 
 When formulating your actions:
 
-1. Use only the following action types: click, drag and drop, hover, press key, select, type.
+1. Use only the following action types: click, drag and drop, hover, press key, select, type, back.
 2. Include the element's tag name in each action.
 3. If text content is present for an element, include it in quotes.
 4. Do not include element IDs in the actions.
--- a/alumnium/agents/planner_prompts/google/system.md
+++ b/alumnium/agents/planner_prompts/google/system.md
@@ -9,7 +9,7 @@ Your goal is to determine a series of actions that will accomplish the task desc
 
 When formulating your actions:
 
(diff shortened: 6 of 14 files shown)


ioachim-hub:
@p0deje 
Actually I am building a "wrapper" on top of `alumnium`.

This is the flow:
1. Have manuals of instructions - procedures of who to do things.
2. Transform them to natural language instructions.
3. Pass them to alumnium executor.

It is more like an extra level of "AI automation" => write tests from manuals.

alumnium/drivers/playwright_driver.py
```diff
@@ -60,6 +60,10 @@ def press_key(self, key: Key):
     def quit(self):
         self.page.close()
 
+    def back(self):
+        self.page.go_back()
+        self.wait_for_page_to_load()
```
p0deje: You don't need to call this explicitly, we'll auto-wait before fetching the accessibility tree next time

alumnium/drivers/selenium_driver.py
```diff
@@ -62,6 +62,10 @@ def press_key(self, key: Key):
     def quit(self):
         self.driver.quit()
 
+    def back(self):
+        self.driver.back()
+        self.wait_for_page_to_load()
```
p0deje: Ditto.

examples/behave/features/environment.py
```diff
@@ -82,6 +82,15 @@ def alumnium(context):
                 'click button "x" near the "this is Al" task',
             ],
         )
+
+    # Add global learning examples for back navigation
+    context.al.learn("Navigate back to the previous page", ["use back navigation"])
```
p0deje: We're not testing this functionality in Behave examples - let's remove.

examples/pytest/conftest.py
```diff
@@ -55,6 +55,15 @@ def al(driver):
     al = Alumni(driver)
     if driver_type == "appium":
         al.driver.delay = 0.1
+
+    # Add global learning examples for back navigation
+    al.learn("Navigate back to the previous page", ["use back navigation"])
```
p0deje: Based on the simplicity of the example, I feel a planner should be able to handle this automatically if you add it to its prompt. https://github.com/alumnium-hq/alumnium/blob/f75b24efa3b79f042d2a15ff21ac073ed71f6ee3/alumnium/agents/planner_prompts/openai/system.md?plain=1#L12.

ioachim-hub:
> @ioachim-hub One of the things I'd like to avoid in Alumnium is to have too many tools that, right now, can be replaced with direct `driver` calls. In my experience, an increasing number of tools flood the context and decrease LLM response determinism. Can you share an example of what you pass down to Alumnium as "executor"?

I agree with you.
Maybe we can close this PR. 
The examples that I have are quite similar to what I wrote on the test. I need to drill from one page to a different one then go back.
I will try to use a different approach for that.

p0deje:
Another option we can explore is to allow configuring which tools are enabled. So we'll support `GoBackTool` but you'll need to enable it:

```python
Alumni(driver, extra_tools=["go_back"])
```

ioachim-hub:
> Another option we can explore is to allow configuring which tools are enabled. So we'll support `GoBackTool` but you'll need to enable it:
> 
> ```python
> Alumni(driver, extra_tools=["go_back"])
> ```

This is an interesting idea.

p0deje:
@ioachim-hub Thank you for the PR. I'm going to merge it as-is, but I'll do follow-up work to keep this tool disabled by default. You will be able to enable it with the following:

```py
from alumnium.tools import NavigateBackTool

Alumni(driver, extra_tools=[NavigateBackTool])
```

ioachim-hub:
> @ioachim-hub Thank you for the PR. I'm going to merge it as-is, but I'll do follow-up work to keep this tool disabled by default. You will be able to enable it with the following:
> 
> ```python
> from alumnium.tools import NavigateBackTool
> 
> Alumni(driver, extra_tools=[NavigateBackTool])
> ```

@p0deje , thank you very much!
If I will have some ideas while playing with this great tool, I will open a discussion.

PR #173, opened 2025-10-12
refactor: move accessibility processing to server

p0deje:
This PR splits the current accessibility tree implementation into two pieces - client and server. The client is responsible for taking raw tree (CDP in JSON, Appium in XML), assigning each element a unique _raw_ ID, and then converting it to an XML string. This raw XML is sent to the server, where it's further processed. It's cleaned up (removes ignored nodes, unnecessary hierarchy, etc). Then, processed XML nodes get their own unique _simplified_ IDs once again, which is necessary to increase the cache hits. The server then sends this processed XML string to agents and maps simplified IDs back to raw IDs. The response with raw IDs is then sent back to the client, which can interact with elements/scope areas based on raw IDs.

Without this PR, each client would have to implement the complex logic of processing accessibility trees. Now, most of the heavy-lifting is centralized on the server.

### DIFF
--- a/alumnium/server/accessibility/base_server_accessibility_tree.py
+++ b/alumnium/server/accessibility/base_server_accessibility_tree.py
@@ -39,11 +39,12 @@ class BaseServerAccessibilityTree(ABC):
         self._simplified_id_counter += 1
         return self._simplified_id_counter
 
+    # Gemini returns ids as floats
     # Llama sometimes returns ids as strings or nested dicts
-    def _extract_id(self, id: int | str | dict) -> int:
+    def _extract_id(self, id: int | float | str | dict) -> int:
         if isinstance(id, int):
             return id
-        elif isinstance(id, str):
+        elif isinstance(id, str) or isinstance(id, float):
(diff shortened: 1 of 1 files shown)


alumnium/server/main.py
```diff
@@ -113,7 +113,8 @@ async def plan_actions(session_id: str, request: PlanRequest):
         raise HTTPException(status_code=status.HTTP_404_NOT_FOUND, detail="Session not found")
 
     try:
-        steps = session.planner_agent.invoke(request.goal, request.accessibility_tree)
+        accessibility_tree = session.process_tree(request.accessibility_tree)
```
sh3pik, line 116: I see that `process_tree` is followed by `to_xml`, should we join them? Like `process_tree` returns tree in xml format

p0deje: In most cases this is correct, but for /actions, /areas, /elements we also use `get_raw_id` and `map_tool_calls_to_raw_id` on the processed accessibility tree.

PR #175, opened 2025-10-14
refactor: monorepo

p0deje:
This PR restructures the project into a monorepo. The idea is to move language-specific code to `packages`:

```sh
❱ ls -l packages
drwxr-xr-x  15 p0deje  staff  480 Oct 14 06:00 python/
drwxr-xr-x  15 p0deje  staff  480 Oct 14 06:00 typescript/
```

Once packages are split, the rest of this PR is just making common tooling work well with the setup:
- VS Code works using `alumnium.code-workspace` where packages form [multi-root-workspaces](https://code.visualstudio.com/docs/editing/workspaces/multi-root-workspaces) 
- Added `poe` to provide simple scripts for `format`, `test`, and others, which allows one to simply execute `poetry poe format` from Makefile
- CI/release pipelines and Makefile targets are updated to switch to `packages/python/` when needed

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,6 +16,9 @@ jobs:
   test:
     name: Test
     runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: packages/python
     steps:
       - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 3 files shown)


sh3pik (approved):

.github/workflows/ci.yml
```diff
@@ -93,10 +100,11 @@ jobs:
         with:
           python-version-file: .tool-versions
       - run: pip install poetry==$(cat .tool-versions | grep poetry | cut -d" " -f2)
+        working-directory: .
```
sh3pik, line 103: I think CI is failing due to it. what's the reason you need to override `working-directory` here? I think all the steps should be in `packages/python`, or I'm missing something

p0deje: We need to do this in the root because `.tool-versions` lives there.
CI fails because it uses the workflow from `master` and not this PR, and expects all Python code in the root. Should be fixed once it's merged.

PR #183, opened 2025-11-02
feat: support github models

p0deje:
Support GitHub Models as a basic provider. It's OpenAI-compatible and allows to easily use Alumnium on GitHub Actions by setting the GitHub token to `OPENAI_API_KEY`.

One other benefit is that it allows us to run basic tests for free on GitHub Actions without the risk of leaking private tokens to PRs. The current free rate limit is 15 RPM.

Sample successful build - https://github.com/alumnium-hq/alumnium/actions/runs/19014246599/job/54299862729.


sh3pik (approved):
👏

PR #203, opened 2025-11-27
test: skip dollar sign when getting amounts

p0deje:
Fixes CI failures by making `get` instruction more concrete and telling Alumnium to ignore money signs ($).


p0deje:
Looks better - https://github.com/alumnium-hq/alumnium/actions/runs/19744557575

Issue #109, opened 2025-05-27
Research thinking models to replace planner agent


p0deje:
Now that we have fast and relatively cheap thinking models, we can try removing the planner agent completely and instead ensure that the actor agent _thinks_ before responding with tool calls. This would significantly simplify the implementation and potentially make #60 less important.

Part of this research is to ensure that CI still passes on low-tier models with the following providers:

- o4-mini for OpenAI
- Gemini 2.5 Flash for Google
- DeepSeek-R1 for DeepSeek

If these models work fine, we can decide what to do with Anthropic, Meta and Ollama providers.

PR #208, opened 2025-12-06
feat!: reasoning models

p0deje:
This is the first attempt to migrate from non-reasoning to reasoning models. It generally improves the stability of the current implementation without massive changes to the current architecture or prompting. 

| Provider | Before | After |
|---------|--------|-------|
| Anthropic | **Claude Haiku 4.5** | **Claude Haiku 4.5** with extending thinking |
| DeepSeek | **DeepSeek-chat** ($0.28/1M input, $0.42/1M output) | **DeepSeek-reasoner** ($0.28/1M input, $0.42/1M output)
| Google | **Gemini 2.0 Flash** ($0.10/1M input, $0.40/1M output) | **Gemini 3 Flash** ($0.50/1M input, $3/1M output) |
| OpenAI | **GPT-4o Mini** ($0.15/1M input, $0.60/1M output) | **GPT-5-nano** ($0.05/1M input, $0.40/1M output) |
| xAI | **Grok-4-fast-non-reasoning** ($0.20/1M input, $0.50/1M output) | **Grok-4-1-fast-reasoning** ($0.20/1M input, $0.50/1M output) |

The overall idea at this point is to simply trade latency for accuracy because models now _think_ before doing something. This is a trade-off that should be compensated for by caching, which we want to improve over time. 

Most models behave the same or better with thinking enabled, though it's worth mentioning that DeepSeek/xAI are following the instructions too strictly, which is a problem in the `check`/`get` API, where we share the same prompt. This can be solved by introducing a separate `VerifierAgent` with its own prompt that is used exclusively in the `check` API.

Some providers are left with non-thinking models:
- GitHub has terrible rate limits for GPT-5, so it's not practical to switch yet
- MistralAI and Ollama need major rework

Let's see how well it behaves in the real world. We can always revert or provide an opt-out from thinking models if necessary. If there is an overall positive trend, we can migrate other providers, introduce `VerifierAgent`, and proceed with #109, removing `PlannerAgent.`

### DIFF
--- a/packages/typescript/examples/mocha/calculator.test.ts
+++ b/packages/typescript/examples/mocha/calculator.test.ts
@@ -1,22 +1,8 @@
 import assert from "assert";
-import { Model, Provider } from "../../src/Model.js";
 import "./globals.js";
 import { navigate } from "./helpers.js";
 
 describe("Calculator", () => {
-  const shouldSkip = () => {
-    if (
-      (process.env.ALUMNIUM_DRIVER || "selenium") === "appium" &&
-      [Provider.ANTHROPIC, Provider.AWS_ANTHROPIC].includes(
-        Model.current.provider
-      )
-    ) {
-      return "Incorrect element is identified";
-    }
-
-    return null;
-  };
-
   before(async () => {
     // Mistral skips '+' button.
     await al.learn("4 / 2 =", [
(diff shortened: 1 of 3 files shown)


packages/python/src/alumnium/server/llm_factory.py
```diff
@@ -25,46 +26,92 @@ def create_llm(model: Model):
...
+                llm = AzureChatOpenAI(
+                    model=model.name,
+                    api_version=azure_openai_api_version,
+                    temperature=0,
+                    seed=1,
+                )
+            else:
+                llm = AzureChatOpenAI(
+                    model=model.name,
+                    api_version="2025-03-01-preview",  # TODO: Remove hardcoding before merge
```
p0deje: Don't forget

PR #212, opened 2025-12-13
feat: improve iOS support

p0deje:


packages/python/src/alumnium/drivers/appium_driver.py
```diff
@@ -156,10 +152,7 @@ def find_element(self, id: int) -> WebElement:
                 props_str = " AND ".join(props)
                 predicate += f" AND {props_str}"
 
-            logger.debug(f"Finding element by predicate: {predicate}")
```
sh3pik, line 159: Are the logs too noisy?

packages/typescript/src/accessibility/XCUITestAccessibilityTree.ts
```diff
@@ -131,6 +137,36 @@ export class XCUITestAccessibilityTree extends BaseAccessibilityTree {
...
+
+    let decoded = str;
+    let previous: string;
+
+    // Keep decoding until no more entities are found (handles double-encoded entities)
+    const MAX_DECODE_ITERATIONS = 10;
+    let iterations = 0;
+    do {
+      previous = decoded;
+      decoded = decoded.replace(
```
sh3pik, line 157: Have you considered [something like](https://docs.python.org/3/library/html.html) `html.unescape(str)`?

p0deje: I did, but there is no such thing in JavaScript unless I pull in extra dependencies like lodash or [unescape](https://www.npmjs.com/package/unescape)

packages/python/src/alumnium/drivers/appium_driver.py
p0deje: I've removed it because we don't log anything in the Android equivalent. Maybe it's worth going through the logs in TypeScript package and add it all over the place.

packages/typescript/src/accessibility/XCUITestAccessibilityTree.ts
sh3pik: @p0deje I missed that it's ts 🤦‍♀️ let me try [decodeURI](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURI) and I'll open a follow up MR if it works

PR #220, opened 2026-01-05
feat!: reflection for `do`/`get`

p0deje:
This PR adds a reflection mechanism that will be used by the upcoming Alumnium MCP. The whole idea is that calling individual actions (do/check/get) should feed back the results to the main agent in the form of an explanation.

1. Calling `do` now returns an object that can be used to reflect on the execution. It has an explanation of the initial plan of actions, all steps that were performed, along with tool calls for each step.
2. Calling `get` now returns a string with an explanation of why the data cannot be retrieved from the page. This allows agents using Alumnium MCP to self-correct based on this feedback. It's a **breaking change** since `get` now never returns `None` for missing data.
3. Calling `check` needs to changes because it already returns explanation strings when assertion passes or throws `AssertionError` with the explanation.

### DIFF
--- a/packages/python/src/alumnium/clients/native_client.py
+++ b/packages/python/src/alumnium/clients/native_client.py
@@ -74,7 +74,7 @@ class NativeClient:
         area = self.session.area_agent.invoke(description, accessibility_tree.to_xml())
         return {"id": accessibility_tree.get_raw_id(area["id"]), "explanation": area["explanation"]}
 
-    def find_element(self, description: str, accessibility_tree: str, area_id: int = None) -> dict:
+    def find_element(self, description: str, accessibility_tree: str) -> dict:
         accessibility_tree = self.session.process_tree(accessibility_tree)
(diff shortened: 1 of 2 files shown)


packages/python/src/alumnium/area.py
```diff
@@ -81,16 +92,16 @@ def get(self, data: str, vision: bool = False) -> Data:
...
-        _, value = self.client.retrieve(
+        explanation, value = self.client.retrieve(
             data,
             self.accessibility_tree.to_str(),
             title=self.driver.title,
             url=self.driver.url,
             screenshot=self.driver.screenshot if vision else None,
         )
-        return value
+        return explanation if value is None else value
```
sh3pik, line 104: @p0deje, this might be confusing, as you will see something like `"There is no such data on the page" doesn't match "34"`. Why do you need it? Does it simplify debugging?

sh3pik: Sorry, I missed "This allows agents using Alumnium MCP to self-correct based on this feedback."

PR #255, opened 2026-02-19
feat: plannerless mode

p0deje:
This allows to disable the planner agent completely by passing `planner=False` to `Alumni()`. Without the planner, `al.do()` goes straight to the actor agent and is capable of executing single-step instructions without preliminary planning phase.

This is desirable in many cases, particularly when you have very detailed step-by-step instructions (e.g. "click 'Next' button") that result in a single-step plans closely matching the instruction itself. For such scenarios, there is little value in paying the performance/cost penalty of planning.

Planner can also be disabled by setting `ALUMNIUM_PLANNER=false` environment variable.

### DIFF
--- a/packages/python/src/alumnium/clients/native_client.py
+++ b/packages/python/src/alumnium/clients/native_client.py
@@ -49,6 +49,7 @@ class NativeClient:
         """
         if not self.session.planner:
             return (goal, [goal])
+
         accessibility_tree = self.session.process_tree(accessibility_tree)
         return self.session.planner_agent.invoke(goal, accessibility_tree.to_xml())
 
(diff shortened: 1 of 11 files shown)


packages/python/src/alumnium/clients/native_client.py
```diff
@@ -41,6 +47,8 @@ def plan_actions(self, goal: str, accessibility_tree: str) -> tuple[str, list[st
         Returns:
             A tuple of (explanation, steps).
         """
+        if not self.session.planner:
+            return (goal, [goal])
```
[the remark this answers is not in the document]
p0deje: One thing that remains unknown is how examples are going to work without planner. Right now they will have no effect, while they should probably be propagated to actor instead.

packages/python/src/alumnium/server/agents/base_agent.py
```diff
@@ -16,6 +18,22 @@
 logger = get_logger(__name__)
 
 
+@dataclass
+class Response:
```
sai-meyyappan, line 22: nit: Response is in base_agent.py, but it’s just a shared data model, not agent logic. So anything outside the agent tree (tests, http_client.py, etc.) has to import it from base_agent, which creates an unnecessary dependency on the base class file.

Probably cleaner to move Response into agents/models.py or agents/types.py. Then both agents and external code can import it from there, and base_agent.py stays focused on agent behavior.

sai-meyyappan (approved):
one nit comment, otherwise lgtm

packages/python/src/alumnium/server/agents/base_agent.py
p0deje: I don't think it's going to be imported directly from the module, this is essentially a private class that is only used by other agents.

PR #270, opened 2026-03-17
feat(cache): remote backends S3 & Redis

kossnocorp:
This PR adds the remote cache feature, allowing the use of Redis or an S3 bucket as the file store backend. This will enable cross-machine cache that allows teams to save time, tokens and also add more consistency.

The current implementation of the fuzzy elements cache relies on continuously going through cache contents (`instruction.json`) that slow down the cache lookups, so it is worth looking into how the cache can be optimized for remote access. That might become a problem with the growing number of cache files.

It's true for both the filesystem and remote backends, so storing cache contents or their parts in memory for a certain period of time (or RAM limit) sounds like a good idea either way. It is possible, as the cache contents are immutable, so once loaded into memory, we can be sure they will never change.


kossnocorp:
Outdated also could be done better

PR #292, opened 2026-04-03
Merge website into `packages/website`

kossnocorp:
This PR merges [the website](https://github.com/alumnium-hq/alumnium) into `packages/website` along with the Git history. It would allow making it easier to create PRs that also adjust documentation and have unified dev & release workflows. It also enables codegen and document reuse: some documents (e.g., `JSDoc` annotations) can be used to generate website pages, and some, like the Contributing Guide, could be reused entirely.

I used the `git subtree` command to make it happen:

```sh
git checkout -c website
git subtree add --prefix=packages/website website main
```

Things to do (can be done before or after the merge):

- Adjust GitHub Pages configuration in the repo to use the subfolder and disable pages in the website repo.
- Merge GitHub Actions workflows.

Integrating it as a full-blown workspace package, making it use common package manager, integrating website tasks, etc., makes sense only after merging #265 so that we don't configure it twice.

It can be done in the `ts` branch too. In that case, we better merge this PR first, rebase `ts` to `main` and tune the thing. It would allow us to land website docs adjustments along with the TypeScript migration, but it also can be done separately.


p0deje:
Merged in b8a0f35a

PR #295, opened 2026-04-07
feat: support alerts across all drivers and languages

### LINKED ISSUE #209: Support alerts
Right now, alerts will block the test execution, and there will be no way around it unless the user calls `driver.switch_to.alert.close()`. Alumnium should automatically add alerts to the accessibility tree so LLM can decide to `click` on its buttons, then seamlessly translate such clicks into respective alert accept/dismiss calls.

This should be implemented across all supported drivers and languages.

ABNclearroute:
## Summary

Resolves #209

- **Detect alerts**: Selenium checks `switch_to.alert`; Playwright captures dialogs via `page.on('dialog')`; Appium handles WebView alerts via `switch_to.alert`
- **Inject synthetic nodes**: When an alert is present, synthetic `alertdialog`, OK (`accept`), and Cancel (`dismiss`) button nodes are added to the accessibility tree so the LLM can see and interact with them
- **Route clicks to alert API**: When the LLM clicks an alert button, the driver intercepts it and calls `alert.accept()` or `alert.dismiss()` instead of trying to find a DOM element

### Driver-specific behavior

| Driver | Behavior |
|---|---|
| **Selenium** | Alert stays open. LLM sees buttons in the tree and clicks to accept/dismiss. |
| **Playwright** | Dialog is auto-accepted (required by Playwright API). LLM sees captured dialog info in the next tree for assertion purposes. |
| **Appium** | Native alerts are already in the page source. WebView alerts are handled via `switch_to.alert`. |

### Files changed (11)

**Python (6 files):**
- `accessibility_element.py` — added `alert_action` field
- `chromium_accessibility_tree.py` — store/read `_alert_action` in XML tree
- `selenium_driver.py` — alert detection, synthetic nodes, click routing, resilient tab-switch decorator
- `playwright_driver.py` — dialog handler, captured dialog nodes, click acknowledgement
- `playwright_async_driver.py` — same as sync, with async-safe dialog acceptance
- `appium_driver.py` — basic alert routing in click

**TypeScript (5 files):**
- `AccessibilityElement.ts` — added `alertAction` field
- `ChromiumAccessibilityTree.ts` — store/read `_alert_action` in XML tree
- `SeleniumDriver.ts` — alert detection, synthetic nodes, click routing, resilient decorator
- `PlaywrightDriver.ts` — dialog handler, captured dialog nodes, click acknowledgement
- `AppiumDriver.ts` — basic alert routing in click

## Test plan

- [x] Selenium: Trigger `alert()` → verify LLM sees alertdialog + OK/Cancel → click OK → alert dismissed
- [x] Selenium: Trigger `confirm()` → click Cancel → confirm returns false
- [x] Playwright: Trigger `alert()` → verify dialog info appears in tree → LLM can assert on text
- [x] Playwright: Trigger `confirm()` → verify auto-accept behavior
- [x] Appium: WebView alert → verify accept/dismiss routing
- [x] Verify normal (non-alert) click/type operations remain unaffected

### DIFF
--- a/packages/python/examples/pytest/alert_test.py
+++ b/packages/python/examples/pytest/alert_test.py
@@ -1,0 +1,13 @@
+def test_alert_accept(al, navigate):
+    navigate("alert.html")
+    al.do("click the Trigger Alert button")
+    al.check("an alert dialog is shown")
+    al.do("accept the alert")
+
+
+def test_confirm_dismiss(al, navigate):
+    navigate("alert.html")
+    al.do("click the Trigger Confirm button")
+    al.check("a confirm dialog is shown")
+    al.do("dismiss the alert")
+    al.check("the result text says Dismissed")
--- a/packages/python/src/alumnium/drivers/playwright_driver.py
+++ b/packages/python/src/alumnium/drivers/playwright_driver.py
@@ -4,7 +4,7 @@ from os import getenv
 from pathlib import Path
 from urllib.parse import urlparse
 
-from playwright.sync_api import Error, Frame, Locator, Page, TimeoutError
+from playwright.sync_api import Dialog, Error, Frame, Locator, Page, TimeoutError
 
 from .. import FULL_PAGE_SCREENSHOT
 from ..accessibility import ChromiumAccessibilityTree
@@ -40,6 +40,8 @@ class PlaywrightDriver(BaseDriver):
         self.page = page
         self.autoswitch_to_new_tab = True
         self.full_page_screenshot = FULL_PAGE_SCREENSHOT
+        self._last_dialog_info: dict | None = None
+        self.page.on("dialog", self._on_dialog)
         self.supported_tools = {
             ClickTool,
             DragAndDropTool,
@@ -57,6 +59,11 @@ class PlaywrightDriver(BaseDriver):
 
     @property
     def accessibility_tree(self) -> ChromiumAccessibilityTree:
+        if self._last_dialog_info:
+            dialog_info = self._last_dialog_info
+            self._last_dialog_info = None
+            return ChromiumAccessibilityTree({"nodes": self._create_dialog_nodes(dialog_info)})
+
         self._wait_for_page_to_load()
 
         # Get frame tree to enumerate all frames (same approach as Selenium)
@@ -144,6 +151,11 @@ class PlaywrightDriver(BaseDriver):
         return ChromiumAccessibilityTree({"nodes": all_nodes})
 
     def click(self, id: int):
+        accessibility_element = self.accessibility_tree.element_by_id(id)
+        if accessibility_element.alert_action:
+            logger.debug(f"Alert action acknowledged: {accessibility_element.alert_action}")
+            return
+
         element = self.find_element(id)
         tag_name = element.evaluate("el => el.tagName")
         if tag_name.lower() == "option":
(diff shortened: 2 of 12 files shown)


p0deje:
Thank you for the PR. A few questions came up as I was looking through it:

1. I think the decision to accept/dismiss the alert should be on the user. The idea was to provide alerts as synthetic nodes in the accessibility tree tools in the actor agent that can be used by a user. This would allow both to inspect the alert with `get/check` and handle it with `do`:

```py
al.do("click on the button that triggers alert")
# accessibility tree now has
# <alert id="1">
#   <button id="2" text="Accept" />
#   <button id="3" text="Dismiss" />
# </alert>
al.check("alert is shown")
al.do("accept alert")  # clicks button with id=2 -> accepts alert
```

2. Is Selenium fully blocked when an alert is present? If not, maybe we can add a `driver.alert` method that returns alert properties and use it when constructing a synthetic node.

ABNclearroute:
@p0deje  Thanks for the feedback! Here's what I've changed:

Selenium — CDP commands work even with an alert open, so we now return the full page tree + alert nodes appended at the end. The alert stays open for the user to accept/dismiss.

Playwright — Auto-accept is unavoidable (Playwright freezes if the dialog handler doesn't resolve immediately). Added clear docstrings explaining this constraint. Dialog info is still exposed in the tree for check()/get().

Node structure — Buttons are now children of alertdialog via childIds:
`<alertdialog name="Alert message">
  <button name="Accept" />
  <button name="Dismiss" />
</alertdialog>`

p0deje:
We've just merged a rewrite of the core into TypeScript - can you please rebase your PR?

Let's also add some tests both in Python and TypeScript. Something simple where Alumnium clicks a button to trigger alert, then checks for alert presence and finally closes it.

packages/python/src/alumnium/drivers/selenium_driver.py
```diff
@@ -349,9 +357,15 @@ def wrapper(self: "SeleniumDriver", *args, **kwargs):
             if not self.autoswitch_to_new_tab:
                 return func(self, *args, **kwargs)
 
-            current_handles = self.driver.window_handles
+            try:
+                current_handles = self.driver.window_handles
+            except Exception:
```
p0deje, line 362: Can we use a concrete exception error (e.g. UnexpectedAlertPresentException)

packages/python/src/alumnium/drivers/selenium_driver.py
```diff
@@ -349,9 +357,15 @@ def wrapper(self: "SeleniumDriver", *args, **kwargs):
...
+                return func(self, *args, **kwargs)
             result = func(self, *args, **kwargs)
-            new_handles = self.driver.window_handles
+            try:
+                new_handles = self.driver.window_handles
+            except Exception:
```
p0deje, line 367: Ditto about UnexpectedAlertPresentException.

packages/python/src/alumnium/drivers/selenium_driver.py
```diff
@@ -366,11 +380,19 @@ def wrapper(self: "SeleniumDriver", *args, **kwargs):
...
+                alert.accept()
+            elif accessibility_element.alert_action == "dismiss":
+                alert.dismiss()
+            return
+
         element = self.find_element(id)
         try:
             ActionChains(self.driver).move_to_element(element).click().perform()
         except ElementNotInteractableException:
-            # Fallback to direct click if ActionChains fails (e.g. for <option> elements)
```
p0deje, line 373: Please revert.

packages/typescript/src/accessibility/ChromiumAccessibilityTree.ts
p0deje, line 1: Please revert formatting changes!

packages/typescript/src/drivers/PlaywrightDriver.ts
p0deje, line 1: Please revert formatting changes!

packages/typescript/src/drivers/SeleniumDriver.ts
p0deje, line 1: Please revert formatting changes!

packages/python/examples/pytest/alert_test.py
p0deje, line 1: The tests need to be disabled for Appium, please see other test to check how it can be done.

packages/typescript/tests/system/alert.test.ts
p0deje, line 1: The tests need to be disabled for Appium, please see other test to check how it can be done.

packages/python/src/alumnium/drivers/playwright_driver.py
```diff
@@ -40,6 +40,8 @@ def __init__(self, page: Page):
         self.page = page
         self.autoswitch_to_new_tab = True
         self.full_page_screenshot = FULL_PAGE_SCREENSHOT
+        self._last_dialog_info: dict | None = None
+        self.page.on("dialog", self._on_dialog)
```
p0deje, line 44: This needs to be applied to all new pages - let's move to `_setup_page_tracking`.

p0deje:
@ABNclearroute would you be able to address my comments and resolve conflicts on this branch?

PR #298, opened 2026-04-13
feat: release stabilization

p0deje:
Bundles multiple fixes that are required to successfully release Python and TypeScript packages with a new `alumnium` binary.

### DIFF
--- a/.github/workflows/ci-typescript.yml
+++ b/.github/workflows/ci-typescript.yml
@@ -50,10 +50,6 @@ jobs:
           - driver: appium-ios
             model: azure_openai
 
-          # TODO: Add support for appium-android to system tests
-          # - driver: appium-android
-          #   model: azure_openai
-
     steps:
       - name: Git Checkout
         uses: actions/checkout@v6
--- a/.github/workflows/ci-typescript.yml
+++ b/.github/workflows/ci-typescript.yml
@@ -108,7 +108,6 @@ jobs:
   build:
     name: Build
     runs-on: ubuntu-latest
-    needs: test-system
 
     steps:
       - name: Git Checkout
--- a/packages/typescript/scripts/build-pip-src.sh
+++ b/packages/typescript/scripts/build-pip-src.sh
@@ -20,6 +20,9 @@ trap 'rm -rf "$tmp_dir"' EXIT
 top_dir="$tmp_dir/alumnium_cli-$VERSION"
 mkdir -p "$top_dir"
 
+# PyPI requires PKG-INFO in the sdist root
+printf '%s' "$PKG_INFO" > "$top_dir/PKG-INFO"
+
 # Copy pip directories to temp, zeroing out binaries
 while IFS= read -r -d '' dir; do
 	cp -r "$dir" "$top_dir/"
--- a/packages/typescript/scripts/build-pip-src.sh
+++ b/packages/typescript/scripts/build-pip-src.sh
@@ -10,16 +10,20 @@ fi
 
 PKG_DIR="$(dirname "${BASH_SOURCE[0]}")/.."
 DIST_DIR="$PKG_DIR/dist"
-SRC_TAR_GZ_NAME="alumnium-cli-$VERSION.tar.gz"
+SRC_TAR_GZ_NAME="alumnium_cli-$VERSION.tar.gz"
 SRC_TAR_GZ_PATH="$DIST_DIR/pip/$SRC_TAR_GZ_NAME"
 
 tmp_dir="$(mktemp -d)"
 trap 'rm -rf "$tmp_dir"' EXIT
 
+# PyPI sdist requires a single top-level directory named {package}-{version}
+top_dir="$tmp_dir/alumnium_cli-$VERSION"
+mkdir -p "$top_dir"
+
 # Copy pip directories to temp, zeroing out binaries
 while IFS= read -r -d '' dir; do
-	cp -r "$dir" "$tmp_dir/"
-	find "$tmp_dir/$(basename "$dir")" -path '*/src/alumnium_cli/alumnium-*' \
+	cp -r "$dir" "$top_dir/"
+	find "$top_dir/$(basename "$dir")" -path '*/src/alumnium_cli/alumnium-*' \
 		-exec sh -c ': > "$1"' _ {} \;
 done < <(find "$DIST_DIR" -maxdepth 1 -type d -name 'pip-alumnium-cli*' -print0)
 
@@ -27,7 +31,7 @@ done < <(find "$DIST_DIR" -maxdepth 1 -type d -name 'pip-alumnium-cli*' -print0)
 tar -czf "$SRC_TAR_GZ_PATH" \
 	--exclude='*.ruff_cache*' \
 	-C "$tmp_dir" \
-	.
+	"alumnium_cli-$VERSION"
 
 if [ "$BUILD_SUBSCRIPT" = "true" ]; then
 	echo "🟢 Pip source tarball: $SRC_TAR_GZ_PATH"
--- a/packages/typescript/scripts/build.ts
+++ b/packages/typescript/scripts/build.ts
@@ -547,12 +547,12 @@ async function buildPipWheel(
 }
 
 function pipWheelFileName(name: string, platformTag: string): string {
-  const version = pipWheelVersion(ALUMNIUM_VERSION);
+  const version = pipVersion(ALUMNIUM_VERSION);
   return `${name.replace(/-/g, "_")}-${version}-py3-none-${platformTag}.whl`;
 }
 
 // Pip and NPM has different rules for alpha versions. NPM allows hyphens (e.g. 0.20.0-alpha.1) while pip doesn't (it expects 0.20.0a1).
-function pipWheelVersion(version: string): string {
+function pipVersion(version: string): string {
   return version.replace(/-alpha\.(\d+)/, "a$1");
 }
 
@@ -620,7 +620,8 @@ function getPyProjectToml(project: PyProject) {
 }
 
 async function generateSourceTarGz() {
-  await $`VERSION=${ALUMNIUM_VERSION} BUILD_SUBSCRIPT=true ./scripts/build-pip-src.sh`;
+  const version = pipVersion(ALUMNIUM_VERSION);
+  await $`VERSION=${version} BUILD_SUBSCRIPT=true ./scripts/build-pip-src.sh`;
 }
 
 async function finalizePip(pipName: string, pipDir: string, tag?: string) {
--- a/packages/typescript/scripts/build.ts
+++ b/packages/typescript/scripts/build.ts
@@ -355,9 +355,7 @@ async function main() {
         distPackageJson.bin = distPackageJson.publishConfig.bin;
 
         TARGET_PLATFORMS.forEach(({ npm }) => {
-          distPackageJson.optionalDependencies[npm.name] = `file:${npm.dir}`;
-          distPackageJson.publishConfig.optionalDependencies[npm.name] =
-            ALUMNIUM_VERSION;
+          distPackageJson.optionalDependencies[npm.name] = ALUMNIUM_VERSION;
         });
 
         const distPackageJsonPath = path.resolve(
--- a/packages/typescript/scripts/build.ts
+++ b/packages/typescript/scripts/build.ts
@@ -188,8 +188,8 @@ const loggerPathPlugin: BunPlugin = {
 
 const seleniumRequireAtomRe = /requireAtom\('([^']+)'.+$/gm;
 
-const depsPatcherPlugin: BunPlugin = {
-  name: "deps-patcher",
+const seleniumAtomsPatcherPlugin: BunPlugin = {
+  name: "selenium-atoms-patcher",
   setup(build) {
     build.onLoad(
       { filter: /selenium-webdriver.+http\.js$/, namespace: "file" },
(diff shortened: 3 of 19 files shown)


.github/workflows/ci-typescript.yml
```diff
@@ -50,10 +50,6 @@ jobs:
           - driver: appium-ios
             model: azure_openai
 
-          # TODO: Add support for appium-android to system tests
-          # - driver: appium-android
-          #   model: azure_openai
-
```
kossnocorp, line 56: It would be great to have it or it's plain impossible?

packages/typescript/scripts/build-pip-src.sh
```diff
@@ -10,24 +10,31 @@ fi
...
 DIST_DIR="$PKG_DIR/dist"
-SRC_TAR_GZ_NAME="alumnium-cli-$VERSION.tar.gz"
+SRC_TAR_GZ_NAME="alumnium_cli-$VERSION.tar.gz"
 SRC_TAR_GZ_PATH="$DIST_DIR/pip/$SRC_TAR_GZ_NAME"
 
 tmp_dir="$(mktemp -d)"
 trap 'rm -rf "$tmp_dir"' EXIT
 
+# PyPI sdist requires a single top-level directory named {package}-{version}
+top_dir="$tmp_dir/alumnium_cli-$VERSION"
```
kossnocorp, line 20: What's top btw?

packages/typescript/scripts/build.ts
```diff
@@ -186,49 +200,71 @@ const loggerPathPlugin: BunPlugin = {
...
+      async (args) => {
+        return {
+          contents: await Bun.file(args.path).bytes(),
+          loader: "file",
+        };
+      },
+    );
+  },
+};
+
```
kossnocorp, line 267: Out of scope, but we need to add tests that check these edge cases. They are very obscure!

packages/typescript/scripts/publish.sh
```diff
@@ -24,7 +24,7 @@ publish_pkg() {
 	[[ -f "$pkg" ]] || return 0
 
 	echo -e "🌀️ Publishing $pkg...\n"
-	PUBLISH_ARGS=("--provenance" "--access public")
+	PUBLISH_ARGS=("--provenance", "--access=public")
```
kossnocorp: It adds `,` to `--provenance`. Bash arrays have no commas.

```suggestion
	PUBLISH_ARGS=("--provenance" "--access=public")
```

packages/typescript/scripts/build.ts
```diff
@@ -355,9 +395,7 @@ async function main() {
         distPackageJson.bin = distPackageJson.publishConfig.bin;
 
         TARGET_PLATFORMS.forEach(({ npm }) => {
-          distPackageJson.optionalDependencies[npm.name] = `file:${npm.dir}`;
-          distPackageJson.publishConfig.optionalDependencies[npm.name] =
-            ALUMNIUM_VERSION;
+          distPackageJson.optionalDependencies[npm.name] = ALUMNIUM_VERSION;
```
kossnocorp, line 398: I feel like removing `publishConfig` forced you to add these lines to the npm tests below, which makes the test more explicit. Doesn't Bun use `publishConfig` when packing?

```
  "pnpm": {
    "overrides": {
      "@alumnium/cli-darwin-arm64": "file:../../../dist/npm-alumnium-cli-darwin-arm64",
      "@alumnium/cli-darwin-x64": "file:../../../dist/npm-alumnium-cli-darwin-x64",
      "@alumnium/cli-linux-arm64": "file:../../../dist/npm-alumnium-cli-linux-arm64",
      "@alumnium/cli-linux-x64": "file:../../../dist/npm-alumnium-cli-linux-x64",
      "@alumnium/cli-windows-arm64": "file:../../../dist/npm-alumnium-cli-windows-arm64",
      "@alumnium/cli-windows-x64": "file:../../../dist/npm-alumnium-cli-windows-x64"
    }
```

packages/typescript/scripts/build.ts
```diff
@@ -748,4 +809,58 @@ function getNpmOs(os: OS) {
...
+> {
+  const seleniumPkgDir = await fs.realpath(
+    path.resolve(PKG_DIR, "node_modules/selenium-webdriver"),
+  );
+  const playwrightCorePkgDir = await fs.realpath(
+    path.resolve(PKG_DIR, "node_modules/playwright-core"),
+  );
+
+  const seleniumAtomPaths = Array.from(
+    new Bun.Glob(path.join(seleniumPkgDir, "lib/atoms/*.js")).scanSync("/"),
```
kossnocorp: It's better to use async even if in practice nothing can get blocked.

packages/typescript/scripts/build.ts
```diff
@@ -748,4 +809,58 @@ function getNpmOs(os: OS) {
...
+
+async function getStandaloneEmbeddedAssets(): Promise<
+  StandaloneEmbeddedAsset[]
+> {
+  const seleniumPkgDir = await fs.realpath(
+    path.resolve(PKG_DIR, "node_modules/selenium-webdriver"),
+  );
+  const playwrightCorePkgDir = await fs.realpath(
+    path.resolve(PKG_DIR, "node_modules/playwright-core"),
+  );
```
kossnocorp, line 836: Do you have to use `fs.realpath` because of symlinks? Worth adding a note.

packages/typescript/scripts/build.ts
```diff
@@ -266,19 +302,23 @@ async function main() {
...
     console.log("\n🌀 Building binaries:\n");
 
+    const standaloneEmbeddedAssetPaths =
+      await prepareStandaloneEmbeddedAssets();
+
     await Promise.all(
       TARGET_PLATFORMS.map(async ({ os, arch, target, binPath }) => {
         const result = await Bun.build({
-          entrypoints: [BIN_SRC_PATH],
+          entrypoints: [BIN_SRC_PATH, ...standaloneEmbeddedAssetPaths],
```
kossnocorp, line 311: How does adding `standaloneEmbeddedAssetPaths` to `entrypoints` work? Worth adding a note. From the code it seems like you're pointing Bun to compile/embed Selenium binaries, but I checked bin sizes and they haven't changed a bit.

packages/typescript/scripts/build.ts
```diff
@@ -620,7 +658,30 @@ function getPyProjectToml(project: PyProject) {
...
+    "Summary: Alumnium CLI",
+    `Author: ${authorNames}`,
+    `Author-email: ${authorEmails}`,
+    "License-Expression: MIT",
+    "Requires-Python: >=3.10, <4.0",
+    "Project-URL: Documentation, https://alumnium.ai/docs/",
+    "Project-URL: Homepage, https://alumnium.ai/",
+    "Project-URL: Issues, https://github.com/alumnium-hq/alumnium/issues",
+    "Project-URL: Repository, https://github.com/alumnium-hq/alumnium",
+  ].join("\n");
```
kossnocorp, line 684: I would use string literal instead to get rid of the noise:

```suggestion
  return `Metadata-Version: 2.4
Name: ${PIP_CLI_PKG_NAME}
Version: ${version}
Summary: Alumnium CLI
Author: ${authorNames}
Author-email: ${authorEmails}
License-Expression: MIT
Requires-Python: >=3.10, <4.0
Project-URL: Documentation, https://alumnium.ai/docs/
Project-URL: Homepage, https://alumnium.ai/
Project-URL: Issues, https://github.com/alumnium-hq/alumnium/issues
Project-URL: Repository, https://github.com/alumnium-hq/alumnium
`;
```

packages/typescript/src/standalone/setupEmbeddedDependencies.ts
```diff
@@ -0,0 +1,217 @@
...
+          name.slice(SELENIUM_ATOM_ASSET_PREFIX.length),
+        ),
+      ),
+    ),
+  );
+
+  const seleniumManagerPath = await extractSeleniumManager(
+    filesByName,
+    extractedDir,
+  );
```
kossnocorp: Can't we put it all in a single `Promise.all`?

```suggestion
  const filesByName = getEmbeddedFilesByName();

  const seleniumAtomNames = Object.keys(filesByName)
    .filter((name) => name.startsWith(SELENIUM_ATOM_ASSET_PREFIX))
    .sort();

  const [seleniumManagerPath] = await Promise.all([
    extractSeleniumManager(filesByName, extractedDir),

    ...seleniumAtomNames.map((name) =>
      writeEmbeddedFile(
        filesByName,
        name,
        path.join(
          seleniumAtomsDir,
          name.slice(SELENIUM_ATOM_ASSET_PREFIX.length),
        ),
      ),
    ),

    writeEmbeddedFile(
      filesByName,
      PLAYWRIGHT_CORE_PACKAGE_JSON_ASSET_NAME,
      playwrightPackageJsonPath,
    ),
  ]);
```

packages/typescript/src/standalone/setupEmbeddedDependencies.ts
```diff
@@ -0,0 +1,217 @@
...
+    _resolveFilename(...args: unknown[]): string;
+  };
+
+  const resolveFilename = moduleWithResolveFilename._resolveFilename;
+
+  moduleWithResolveFilename._resolveFilename = function (...args: unknown[]) {
+    const [request] = args;
+
+    if (typeof request === "string") {
+      if (request.startsWith("./atoms/")) {
```
kossnocorp, line 154: I see that you check `"\\playwright-core\\package.json"` below. I assume that's for Windows. Should it use `path.sep` here too?

packages/typescript/src/standalone/setupEmbeddedDependencies.ts
```diff
@@ -0,0 +1,217 @@
...
+        return path.join(
+          paths.seleniumAtomsDir,
+          request.slice("./atoms/".length),
```
kossnocorp, line 157: Same here.

packages/typescript/src/standalone/setupEmbeddedDependencies.ts
```diff
@@ -0,0 +1,217 @@
...
+        );
+      }
+
+      if (request === "../../../package.json") {
```
kossnocorp: And here.

packages/typescript/src/standalone/setupEmbeddedDependencies.ts
```diff
@@ -0,0 +1,217 @@
...
+        return paths.playwrightPackageJsonPath;
+      }
+
+      if (
+        request.endsWith("/playwright-core/package.json") ||
+        request.endsWith("\\playwright-core\\package.json")
```
kossnocorp: It would be easier to use `join` here `path.sep + path.join("playwright-core", "package.json"`. I feel like the leading `path.sep` can be trimmed though.

packages/typescript/src/standalone/setupEmbeddedDependencies.ts
```diff
@@ -0,0 +1,217 @@
...
+    throw new Error(`Missing embedded dependency asset: ${embeddedFileName}`);
+  }
+
+  await fs.mkdir(path.dirname(targetPath), { recursive: true });
+  await fs.writeFile(targetPath, await file.bytes());
+
+  if (mode) {
+    await fs.chmod(targetPath, mode);
+  }
+}
```
kossnocorp, line 206: That's some proper dark magic, respect 🫡

kossnocorp (changes requested):
The only critical-looking things are:

1. The OS-specific separator usage inside `setupEmbeddedDependencies.ts`, but it's worth checking if Bun doesn't normalize them into URLs.
2. Extra comma in a Bash array.

Everything else is optional.

.github/workflows/ci-typescript.yml
p0deje: It's possible, but either I need to move native app tests from Python to TypeScript, or figure out how to make Chrome browser work on Android.

packages/typescript/scripts/publish.sh
p0deje: Good catch!

packages/typescript/scripts/build.ts
p0deje: The problem was that `publishConfig.optionalDependencies` somehow was completely ignored. When I pushed the built package to NPM and tried using it, I was getting errors because it tried installing from `file:`

```
$ npx alumnium@0.20.0-alpha.3
node:internal/modules/package_json_reader:316
  throw new ERR_MODULE_NOT_FOUND(packageName, fileURLToPath(base), null);
        ^

Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@alumnium/cli-darwin-arm64' imported from /Users/p0deje/node_modules/alumnium/src/cli/binWrapper.js
    at Object.getPackageJSONURL (node:internal/modules/package_json_reader:316:9)
    at packageResolve (node:internal/modules/esm/resolve:768:81)
    at moduleResolve (node:internal/modules/esm/resolve:859:18)
    at defaultResolve (node:internal/modules/esm/resolve:991:11)
    at #cachedDefaultResolve (node:internal/modules/esm/loader:712:20)
    at #resolveAndMaybeBlockOnLoaderThread (node:internal/modules/esm/loader:729:38)
    at ModuleLoader.resolveSync (node:internal/modules/esm/loader:758:52)
    at #resolve (node:internal/modules/esm/loader:694:17)
    at ModuleLoader.getOrCreateModuleJob (node:internal/modules/esm/loader:614:35)
    at onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:643:32) {
  code: 'ERR_MODULE_NOT_FOUND'
}
```

packages/typescript/scripts/build.ts
p0deje: I've found that if you add static files, they are considered "imported" and will be added to the binary and Bun will know how to resolve them later in runtime when they are actually imported. https://bun.sh/blog/bun-v1.1.25#embedding-files-in-standalone-executables-without-importing

I feel like theses files were always there, but Bun could explain how to resolve path to them in runtime.

packages/typescript/scripts/build.ts
p0deje: I'll keep array, don't want to mess with indentation.

packages/typescript/src/standalone/setupEmbeddedDependencies.ts
p0deje: I actually didn't test Windows support at all, let me revert it now and then test+fix separately.

packages/typescript/src/standalone/setupEmbeddedDependencies.ts
p0deje: I wish Bun had some sort of virtual filesystem so that we don't have to monkey patch! https://github.com/oven-sh/bun/issues/5445

packages/typescript/scripts/build-pip-src.sh
p0deje: Screenshot 2026-04-14 at 07 38 42

packages/typescript/scripts/build.ts
p0deje: I agree, though it's very hard to test those w/o publishing the package!

packages/typescript/src/standalone/setupEmbeddedDependencies.ts
kossnocorp: It actually DOES have it https://bun.com/docs/bundler#mix-disk-and-virtual-files

PR #299, opened 2026-04-14
chore: set up types scan + make LangChain usage more type safe

kossnocorp:
This pull request introduces types scan utilities that incrementally analyze data structures and write resulting JSON to the file system. Below is the contributing guide that explains how it works in great detail.

The PR also utilizes the generated schemas and updates `invokeChain`'s result in `BaseAgent` to be more type-safe and accurate, as well as reworks how stats are computed and also adds new `cache_creation`, `cache_read` and `reasoning` that will help optimize LLM going forward.

<details>

<summary>See the Types Scan contributor document</summary>

---

This document describes the `scanTypes` defined in [`packages/typescript/src/utils/typesScan.ts`](../../packages/typescript/src/utils/typesScan.ts), and its counterpart script [`packages/typescript/scripts/generate-types-scan-schema.ts`](../../packages/typescript/scripts/generate-types-scan-schema.ts).

## Quick Start

The idea of the types scan is to assist in adding type safety and provide an insight into the shape of weakly typed data structures. It is important given that Alumnium has to work with numerous ever-changing 3rd-party APIs. New models and project features require additional detail in the existing schemas, so it will be instrumental going forward.

We initially built it to type LangChain generations but can use it out of the context of LLM APIs.

To use it, set the `ALUMNIUM_DEV_DATA_TYPES_SCAN=true` env var (any non-empty value) and call the `scanTypes` function, passing the current module URL and ID, e.g.,:

```ts
const stored = serializeGeneration(generation);
scanTypes(import.meta.url, "serialized", stored);
// ...
```

Once the function is executed (e.g., when running system tests), it creates a JSON file next to the module with the data types information. For example, when called from `Lchain.ts` with the arguments above, it would create `Lchain.serialized.types-scan.json` with content looking like that:

```json
[
  {
    "kind": "object",
    "properties": {
      "text": {
        "kind": "field",
        "optional": false,
        "type": [
          {
            "kind": "string",
            "values": ["Hello", "cruel", "world"]
          }
        ]
      }
    }
  }
]
```

See a types scan example: [`packages/typescript/src/llm/Lchain.serialized.types-scan.json`](../../packages/typescript/src/llm/Lchain.serialized.types-scan.json).

Each `scanTypes` reads the existing schema and extends it with the passed data types. The more examples it sees, the more detailed the `types-scan.json` file becomes.

It doesn't guarantee the 100% match and can only be as detailed as the diverse passed data is.

While processing a new data structure or its variant, it produces literals for every primitive data type, so to get the most out of it, a contributor must manually edit the `types-scan.json` to shape the schema.

See [Shaping `types-scan.json`](#shaping-types-scanjson) for the detailed reference.

#### Generating Schema

To generate Zod schema code from a `type-scan.json` file, use one of the following tasks:

- [`//packages/typescript:generate/types-scan-schema`](#module-mode): Generate a Zod schema module with schema and types exported on the root level.
- [`//packages/typescript:generate/types-scan-schema:class`](#class-mode): Generate a Zod schema abstract class module.

The output can be used to codegen a source file or to cherry-pick needed schemas.

##### Module Mode

Use the `//packages/typescript:generate/types-scan-schema` task to generate a module that exports Zod schemas and inferred types on the root level:

```sh
mise //packages/typescript:generate/types-scan-schema ./src/llm/Lchain.serialized.types-scan.json
```

It outputs such TypeScript code:

```ts
import z from "zod";

export const MessageThinking = z.object({
  type: z.literal("thinking"),
  thinking: z.string(),
  signature: z.string().optional(),
});

export type MessageThinking = z.infer<typeof MessageThinking>;

export const MessageText = z.object({
  type: z.literal("text"),
  text: z.string(),
  annotations: z.array(z.unknown()).optional(),
});

export type MessageText = z.infer<typeof MessageText>;

// ...
```

##### Class mode

Use the `//packages/typescript:generate/types-scan-schema:class` task to generate a module that exports Zod schemas and inferred types wrapped into `abstract class` along with a `namespace`.

```sh
mise //packages/typescript:generate/types-scan-schema:class ./src/llm/Lchain.serialized.types-scan.json LchainSchema
```

> The second argument (`LchainSchema`) specifies the class name to use.

It outputs such TypeScript code:

```ts
import z from "zod";

export abstract class LchainScan {
  static MessageThinking = z.object({
    type: z.literal("thinking"),
    thinking: z.string(),
    signature: z.string().optional(),
  });

  static MessageText = z.object({
    type: z.literal("text"),
    text: z.string(),
    annotations: z.array(z.unknown()).optional(),
  });

  static MessageDataContent = z.discriminatedUnion("type", [
    this.MessageThinking,
    this.MessageText,
  ]);
}

export namespace LchainScan {
  export type MessageThinking = z.infer<typeof LchainScan.MessageThinking>;

  export type MessageText = z.infer<typeof LchainScan.MessageText>;

  export type MessageDataContent = z.infer<
    typeof LchainScan.MessageDataContent
  >;
}
```

## How to Use

After adding a new model support provider or using a new LLM feature, make sure to run a system test with `ALUMNIUM_DEV_DATA_TYPES_SCAN` to update the existing `types-scan.json` files and also instrument any new data sources if needed.

While it can add extra unrelated work (or even irrelevant information), we advise keeping `ALUMNIUM_DEV_DATA_TYPES_SCAN` on and submitting PRs with any updates to the `types-scan.json` files and corresponding schemas.

To capture as much detail as possible, make sure to run the instrumented code multiple times and use different models when relevant. In between runs, make sure to inspect the updated `types-scan.json` and shape it so it is correct. See [Shaping `types-scan.json`](#shaping-types-scanjson) for a detailed reference of various shaping methods.

It is also a good idea to clear the LLM cache before attempting to capture data type updates:

```sh
mise //:clear/cache
```

### Pinning Primitives

The `types-scan.json` is supposed to be manually edited by the contributor using it to shape the schema.

The initial run on a new data structure or its variation generates literals for all primitive data types, which is undesirable in most cases, e.g., semantically, `data.message` is any `string`:

```json
[
  {
    "kind": "object",
    "properties": {
      "text": {
        "kind": "field",
        "optional": false,
        "type": [
          {
            "kind": "string",
            "values": ["Hello", "cruel", "world"]
          }
        ]
      }
    }
  }
]
```

To pin non-literal primitives, replace `"values": [...]` with `"literal": false`. On the next run, the `scanTypes` will recognize it as a generic primitive and will stop adding new values to it.

It turns `z.union([z.literal("Hello"), z.literal("cruel"), z.literal("world")])` into more accurate `z.string()`.

### Pinning Records

Some objects (i.e., LLM tool input that varies from tool to tool) don't need to be detailed, e.g.,:

```json
[
  {
    "kind": "object",
    "properties": {
      "input": {
        "kind": "field",
        "optional": false,
        "type": [
          {
            "kind": "object",
            "properties": {
              "goal": {
                "kind": "field",
                "optional": true,
                "type": [
                  {
                    "kind": "string",
                    "values": ["Goal 1", "Goal 2", "Goal 3"]
                  }
                ]
              }
            }
          }
        ]
      }
    }
  }
]
```

To make the `scanTypes` ignore its shape, set the `kind` property to `"record"`:

```json
[
  {
    "kind": "object",
    "properties": {
      "input": {
        "kind": "field",
        "optional": false,
        "type": [
          {
            "kind": "record"
          }
        ]
      }
    }
  }
]
```

On the next run the `scanTypes` will ignore its internal shape and produce `z.record(z.string(), z.unknown())`.

To make the `scanTypes` keep collecting internal types while ignoring its keys, set the record type:

```json
[
  {
    "kind": "object",
    "properties": {
      "map": {
        "kind": "field",
        "optional": false,
        "type": [
          {
            "kind": "record",
            "type": [
              {
                "kind": "string",
                "literal": false
              }
            ]
          }
        ]
      }
    }
  }
]
```

> Note that `type` must always be an array for any data type for consistency between singular and union types.

### Defining Discriminated Unions

Some data types must be represented as unions rather than objects with variable shape. To make the `scanTypes` collect discriminated union types, assign the `discriminator` property to the field to use as the discriminator.

**Important**: Make sure to run the instrumented code after assigning a discriminator to allow the `scanTypes` update the affected `types-scan.json` files.

For example, an object that can be of two shapes:

```json
{
  "type": "text",
  "text": "Hello, world!"
}
```

...and:

```json
{
  "type": "image",
  "base64": "data:image/jpeg;base64..."
}
```

The initial run would produce:

```json
[
  {
    "kind": "object",
    "properties": {
      "type": {
        "kind": "field",
        "optional": false,
        "type": [
          {
            "kind": "string",
            "values": ["text", "image"]
          }
        ]
      },
      "text": {
        "kind": "field",
        "optional": true,
        "type": [
          {
            "kind": "string",
            "values": ["Hello, world!"]
          }
        ]
      },
      "base64": {
        "kind": "field",
        "optional": true,
        "type": [
          {
            "kind": "string",
            "values": ["data:image/jpeg;base64..."]
          }
        ]
      }
    }
  }
]
```

To turn it into a discriminated union, add `discriminator`:

```jsonc
[
  {
    "kind": "object",
    "discriminator": "type",
    "properties": {
      // ...
    },
  },
]
```

On the next run with the same data, the `scanTypes` will update it and produce a union instead:

```json
[
  {
    "kind": "object",
    "discriminator": "type",
    "properties": [
      {
        "kind": "object",
        "properties": {
          "type": {
            "kind": "field",
            "optional": false,
            "type": [
              {
                "kind": "string",
                "values": ["text"]
              }
            ],
            "discriminator": true
          },
          "text": {
            "kind": "field",
            "optional": false,
            "type": [
              {
                "kind": "string",
                "values": ["Hello, world!"]
              }
            ]
          }
        }
      },
      {
        "kind": "object",
        "properties": {
          "type": {
            "kind": "field",
            "optional": false,
            "type": [
              {
                "kind": "string",
                "values": ["image"]
              }
            ],
            "discriminator": true
          },
          "base64": {
            "kind": "field",
            "optional": false,
            "type": [
              {
                "kind": "string",
                "values": ["data:image/jpeg;base64..."]
              }
            ]
          }
        }
      }
    ]
  }
]
```

The resulting Zod schema will accurately represent the shape of the object:

```ts
import z from "zod";

const MessageText = z.object({ type: z.literal("text"), text: z.string() });

const MessageImage = z.object({ type: z.literal("image"), base64: z.string() });

const Message = z.discriminatedUnion("type", [MessageText, MessageImage]);
```

> For this example, we also assigned `name` and set `literal` to `false`, but to avoid confusion the examples above doesn't reflect it.

### Naming Schemas

By default, when calling the `//packages/typescript:generate/types-scan-schema` task, it derives names from file and node parents, resulting in unique but often verbose names, e.g., `StoredGenerationMessageDataAdditionalKwargsToolCallsItem`.

To customize it, assign the `name` property to a node to use for the node schema representation and all its children, e.g., (on `data.message.data.additional_kwargs`):

```jsonc
[
  {
    "name": "MessageKwargs",
    "kind": "object",
    "properties": {
      "tool_calls": {
        "kind": "field",
        "optional": false,
        "type": [
          // ...
        ],
      },
    },
  },
]
```

It will ignore the parents' prefix and assign more friendly names, e.g., `MessageKwargs` instead of `StoredGenerationMessageDataAdditionalKwargs` and `MessageKwargsToolCallsItem` instead of `StoredGenerationMessageDataAdditionalKwargsToolCallsItem`.

</details>


packages/typescript/src/Model.ts
```diff
@@ -22,7 +22,7 @@ export class ModelName {
     github: "gpt-4o-mini",
     google: "gemini-3.1-flash-lite-preview",
     mistralai: "mistral-medium-2505",
-    ollama: "mistral-small3.1",
+    ollama: "mistral-small3.1:24b",
```
p0deje, line 25: Is it really needed? Seems to work fine on Python without the suffix.

packages/typescript/tests/system/calculator.test.ts
```diff
@@ -20,7 +20,7 @@ describe("Calculator", () => {
     };
   });
 
-  it("addition", async ({ expect, setup }) => {
+  it("addition", { tags: ["cache"] }, async ({ expect, setup }) => {
```
p0deje: I suggest we don't handpick a single test to run - any system test can be used for this purpose.

p0deje (approved):

packages/typescript/src/server/agents/BaseAgent.additional-kwargs.types-scan.json
p0deje, line 1: Is this still needed?

packages/typescript/src/server/agents/BaseAgent.ts
```diff
@@ -254,82 +249,90 @@ export class BaseAgent {
...
   }
 
-  #extractReasoningFromAdditional(additional: object): string | null {
-    if (!additional || typeof additional !== "object") return null;
-    const value = (additional as Record<string, unknown>)["reasoning_content"];
-    return typeof value === "string" && value ? value : null;
+  #extractReasoningFromAdditional(
+    additional: AIMessage["additional_kwargs"],
+  ): string | null {
+    scanTypes({
```
p0deje: I think this should be removed now.

packages/typescript/src/server/LlmFactory.ts
```diff
@@ -267,7 +267,7 @@ export class LlmFactory {
   static createOllamaLlm(model: Model, cache: BaseCache): BaseChatModel {
     logger.debug(`Creating Ollama LLM with model ${model.name}`);
 
-    const baseUrl = process.env.ALUMNIUM_OLLAMA_URL;
+    const baseUrl = process.env.OLLAMA_HOST || process.env.ALUMNIUM_OLLAMA_URL;
```
p0deje, line 281: Please add it to websites/docs/src/content/docs/docs/reference.md

packages/typescript/pnpm-lock.yaml
p0deje, line 1: Is this intentional?

mise.toml
```diff
@@ -99,8 +116,13 @@ depends = [
 depends = ["types/ts", "//packages/python:types"]
 
 [tasks.lint]
```
p0deje, line 119: WDYT if we merge format + lint into a single task?

kossnocorp: I think these are very separate steps that use different tools with different implications. I.e., format can always be applied, while linter often can't safely resolve issues. If you want a unified task that can be used as a hook, then a unifying task would be useful (like `clean` or `neaten`).

packages/typescript/src/server/agents/BaseAgent.ts
kossnocorp: It's replaced by a dummy function during the build, so it has no runtime overhead. I think it will be helpful to have it around so we can scan again when a new model is introduced.

packages/typescript/src/server/agents/BaseAgent.additional-kwargs.types-scan.json
kossnocorp: Yeah, the types-scan files are manually adjusted (i.e., assigned names, set non-literal primitives, etc.), so we need to keep them for when we introduce new models or existing models get updated. Maybe we can have a centralized `./packages/typescript/<dir>` where we store them if you wish not to see them around source code.

packages/typescript/src/Model.ts
kossnocorp: It was needed and didn't work with LangChain SDK, but it's irrelevant now, yeah.

packages/typescript/tests/system/calculator.test.ts
kossnocorp: I don't know if Vitest can run a random test. Besides, some are more flaky than others. What's harm in it?

packages/typescript/src/server/agents/BaseAgent.ts
p0deje: This feels surgical. Why have it in `extractReasoningFromAdditional` but not in `extractReasoningFromContent`? 

My understanding of the whole PR is that there is a _single_ JSON schema that defines all types for all LangChain invokations, regardless of the place it's being called from. So when a new model is added, I don't need to add `scanTypes` to some specific places in code. Instead, I just run all system tests with typescanning turned on. It automatically updates the JSON schema, which I can then edit manually and generate types from. Do I miss something?

packages/typescript/src/server/agents/BaseAgent.additional-kwargs.types-scan.json
p0deje: It's fine to have it in code, but I was under the impression we'll have just _one_ JSON with all LangChain schema.

PR #300, opened 2026-04-14
feat(mcp): return all response as JSON

p0deje:
This PR ensures that all MCP tool responses are JSON. This allows processing the responses via `jq` in Claude Code `PostToolUses` hooks (e.g., to extract `artifacts_store`) after the driver is stopped and automatically generate report based on the artifacts. 

In addition to that, the `capabilities` input in `start_driver` can be read directly from a file rather than constructed by an agent.

### DIFF
--- a/packages/typescript/scripts/build.ts
+++ b/packages/typescript/scripts/build.ts
@@ -336,6 +336,8 @@ async function main() {
         console.log(`🟢 ${target} (${cwdRelPath(binPath)})`);
       }),
     );
+
+    await cleanUpDir(STANDALONE_EMBEDDED_ASSETS_DIR);
   }
 
   //#endregion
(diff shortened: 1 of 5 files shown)


packages/typescript/scripts/build.ts
```diff
@@ -336,6 +336,8 @@ async function main() {
         console.log(`🟢 ${target} (${cwdRelPath(binPath)})`);
       }),
     );
+
+    await cleanUpDir(STANDALONE_EMBEDDED_ASSETS_DIR);
```
kossnocorp, line 340: Better to do it BEFORE the build, as if it's really needed, it might break if a build script version fails for some reason before reaching this point.

p0deje: It's already done before, but the problem was that it left some files in `tmp/` folder that are then being linted during `:format:check` task.

PR #304, opened 2026-04-15
fix(ts): playwright loading on windows

p0deje:


kossnocorp (approved):
Tested both server and MCP commands on Windows! With a small Selenium logging fix, it now works flawlessly 👏

PR #318, opened 2026-04-21
refactor: revamp install scripts

kossnocorp:
This PR makes several changes improving the install scripts behavior:

- Make them static, so they are always in the repo in a final form.
- Use GitHub API to fetch latest version or optionally accept it via `VERSION` env var.
- Fix Windows version:
  - Fix `Join-Path` usage.
  - Prefer using `curl.exe` (that comes with Windows since Windows 10)
  - Fix `Invoke-WebRequest` indicator overlaying terminal history.
- Format both scripts code.


p0deje:
Cherry-picked PS1 fixes.

PR #320, opened 2026-04-24
feat: MCP scenarios

kossnocorp:
🚧 Work in progress!

This PR adds MCP scenarios feature, adding a bunch of new MCP tools that allow coding agents to record and play scenarios from a free-form test description. It is working and tested on simple cases, but UX is far from optimal. Also, prompts need a lot of tuning to make agents recognize the tools as something to use without explicit mentioning.

Try on the following prompts (in different agent sessions).

**Recording**:

````
I want you to record a scenario. Remember, first you have to start a recording and only then start a driver so it is recorded as well.

Here's the scenario:

```
Test https://seleniumbase.io/apps/calculator and check the following calculations for correctness:

2 + 2 = ?
10 / 5 = ?
0 / 0 = ? 
```
````

**Playing**:

````
I want you to play a recorded scenario. Use the `lookup_scenario` tool and then play it:

```
Test https://seleniumbase.io/apps/calculator and check the following calculations for correctness:

2 + 2 = ?
10 / 5 = ?
0 / 0 = ? 
```
````

It generates a scenario that looks like this:

<details>

<summary><code>.alumnium/mcp/scenarios/v8pdr2Is5SIO4I7ZJ4Lzm.json</code></summary>

```json
{
  "kind": "scenario",
  "id": "v8pdr2Is5SIO4I7ZJ4Lzm",
  "text": "Test https://seleniumbase.io/apps/calculator and check the following calculations for correctness:\n\n2 + 2 = ?\n10 / 5 = ?\n0 / 0 = ?",
  "steps": [
    {
      "kind": "step",
      "id": "MhyM8OXl0aPvuvOuDg9CN",
      "toolName": "start",
      "maskedInput": {
        "capabilities": "{\"platformName\":\"chrome\",\"alumnium:options\":{\"headless\":true}}"
      },
      "maskedOutput": {
        "id": "<MASKED_0>",
        "driver": "selenium",
        "model": "azure_openai/gpt-5-nano",
        "platform_name": "chrome"
      }
    },
    {
      "kind": "step",
      "id": "bzWwax1wqavjk9kRisRms",
      "toolName": "do",
      "maskedInput": {
        "id": "<MASKED_0>",
        "goal": "Navigate to URL https://seleniumbase.io/apps/calculator"
      },
      "maskedOutput": {
[...]

PR #328, opened 2026-04-30
feat: add telemetry to track and analyze Alumnium performance

kossnocorp:
This feature was extracted from #320, where, to make the scenarios feature viable, I had to analyze the agents and cache performance.

Since it required instrumentation of the code, with @p0deje we decided to extract and merge it as soon as possible to avoid further conflicts.

---

The PR introduces the following changes:

1. Complete rewrite of the logger to make it more reliable and easier to understand its internals.
   
 An attempt to use an AI agent to set up the OpenTelemetry integration with LogTape made a giant mess and ended up with several configuration strategies depending on passed env vars. The root issue was that we used to synchronously configure LogType while modules were loading and calling `getLogger` on the root module level.

  Now, LogTape gets configured lazily on the first log call, allowing us to change configuration (i.e., setting the level in tests or log path in MCP) and only once apply it when the logger is needed.

2. Introduction of the tracer module that allows you to instrument code with spans and events to analyze its performance. It uses the OpenTelemetry SDKs and allows us and Alumnium users to utilize its massive ecosystem. I also instrumented all core Alumnium components, in some cases, like the Selenium driver, adding more detail.

3. Added a local Grafana stack bundled as a Docker image with OpenTelemetry Collector, Prometheus, Tempo, Loki, and Pyroscope.

It also re-introduces global `.vscode/settings.json` so it would work without using the VS Code workspace, and makes Oxfmt the default formatter as well as formats all the files with it.

## Testing

To enable tracing, pass the `ALUMNIUM_TRACE=true` env var to a command (MCP/server/tests) after starting the Grafana stack:

```bash
mise //telemetry:docker/up
```

Grafana will be available at http://localhost:3000.

If everything's working correctly, you will see traces on the [Drilldown -> Traces](http://localhost:3000/a/grafana-exploretraces-app/explore) page.

## Learn More

See the `docs/contributing/telemetry.md` document for more details.


p0deje (approved):
Amazing! We'll need to add more spans and instrumentation, but that's a solid foundation and already gives many insights of MCP server execution.

p0deje (approved):

PR #330, opened 2026-04-30
perf: experimental waiter script optimization

kossnocorp:
🚧 Work in progress

---

This PR optimizes the waiter script that reduces its duration from ~500-800ms on each `waitForPageToLoad` call. Given how often it's used, it might cut dozens of seconds from just a few steps scenario.

It also adds more advanced tracer instrumentation, logs & telemetry preservation in tests, and a few more goodies.


kossnocorp:
Superseded by https://github.com/alumnium-hq/alumnium/pull/432

PR #331, opened 2026-05-01
fix(python): add missing -> None return type hints in alumni.py

devimano2011:
## Problem

Three methods in `alumni.py` are missing `-> None` return type annotations, inconsistent with the rest of the class where all other methods have explicit return type hints:

- `quit(self):`
- - `learn(self, goal: str, actions: list[str]):`
- - `clear_learn_examples(self):`
## Fix

Add `-> None` return type annotations to these three methods to match the style used by `do()`, `check()`, `get()`, `find()`, `area()`, and `stats()`.

### DIFF
--- a/packages/python/src/alumnium/alumni.py
+++ b/packages/python/src/alumnium/alumni.py
@@ -78,7 +78,7 @@ class Alumni:
 
         self.cache = Cache(self.client)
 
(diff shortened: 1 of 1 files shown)


p0deje:
Thanks for the PRs!

PR #335, opened 2026-05-01
feat(mcp): auto-install playwright browser

p0deje:
Automatically installs Playwright browsers (Chromium, chromium-headless-shell, and FFmpeg) on the first start tool call when `ALUMNIUM_DRIVER=playwright` is set, so users no longer need to run `npx playwright install` manually. Because the compiled Bun binary bundles its own copy of playwright-core, using `npx` from the user's shell risks a version mismatch; instead, the install is driven programmatically using the same internal registry that backs npx playwright install, pinned to the bundled version. In the compiled binary, playwright-core downloads browsers by forking `oopDownloadBrowserMain.js`, but that script's relative imports only exist inside `$bunfs` and can't be executed directly — so `oopDownloadBrowserMain.js` is now pre-bundled at Alumnium build time into a self-contained CJS file with all imports inlined, embedded as an asset, extracted to a temp directory at runtime, and child_process.fork is briefly monkey-patched to redirect calls to it (with `BUN_BE_BUN=1` so the binary behaves as the Bun CLI). In dev mode the patch is skipped entirely and Playwright's normal fork path runs unchanged.

### DIFF
--- a/packages/typescript/src/mcp/mcpDrivers.ts
+++ b/packages/typescript/src/mcp/mcpDrivers.ts
@@ -5,6 +5,7 @@
 import type { BrowserContext, Page } from "playwright-core";
 import { chromium } from "playwright-core";
 import { Builder, type WebDriver } from "selenium-webdriver";
+import { ensurePlaywrightChromiumInstalled } from "../standalone/installPlaywrightBrowsers.ts";
 import { Options } from "selenium-webdriver/chrome.js";
 import {
   remote as remoteWebdriverio,
@@ -85,6 +86,7 @@ export async function createPlaywrightDriver(
     logger.debug("Setting extra HTTP headers: {headers}", { headers });
   }
 
+  await ensurePlaywrightChromiumInstalled();
   const videosDir = await artifactsStore.ensureDir("videos");
 
   let context: BrowserContext;
--- a/packages/typescript/src/standalone/installPlaywrightBrowsers.ts
+++ b/packages/typescript/src/standalone/installPlaywrightBrowsers.ts
@@ -1,0 +1,171 @@
+// Auto-installs Playwright browsers (Chromium + FFmpeg) on first use.
+//
+// In dev mode `installBrowsersForNpmInstall` works as-is because all
+// node_modules are on disk. In the compiled Bun single-file binary it needs
+// extra wiring: playwright-core downloads browsers by forking itself via
+// child_process.fork to run oopDownloadBrowserMain.js. That script lives
+// inside $bunfs and its relative imports (manualPromise, network, zipBundle,
+// fileUtils) aren't available on disk. To fix this we:
+//
+//   1. Pre-bundle oopDownloadBrowserMain.js at Alumnium build time into a
+//      self-contained CJS file with all imports inlined (scripts/build.ts).
+//   2. Embed the bundle as an asset and extract it to a temp dir at runtime
+//      (setupEmbeddedDependencies.ts).
+//   3. Monkey-patch child_process.fork to redirect calls targeting
+//      oopDownloadBrowserMain.js to the extracted bundle.
+//   4. Set BUN_BE_BUN=1 on the fork so process.execPath (the alumnium binary)
+//      behaves as the Bun CLI and can execute the script.
+//
+// The patch is applied only while installBrowsersForNpmInstall is running and
+// removed immediately after, leaving fork unmodified for all other callers.
+import childProcess from "node:child_process";
+import fs from "node:fs";
+import path from "node:path";
+import { Logger } from "../telemetry/Logger.ts";
+import { isSingleFileExecutable } from "../bundle.ts";
+
+const logger = Logger.get(import.meta.url);
+
+let installPromise: Promise<void> | undefined;
+
+// Memoize the promise so concurrent `start` tool calls don't race each other
+// into parallel downloads. The promise is cleared on failure so the next call
+// can retry.
+export function ensurePlaywrightChromiumInstalled(): Promise<void> {
+  installPromise ??= installChromiumIfNeeded().catch((err) => {
+    installPromise = undefined;
+    throw err;
+  });
+  return installPromise;
+}
+
+async function installChromiumIfNeeded() {
+  // String literal dynamic import so Bun bundles playwright-core's registry
+  // at build time from alumnium's own node_modules. A variable import()
+  // is not bundled and fails at runtime in the binary with "Module not found".
+  //
+  // playwright-core/lib/server/registry/index is the same module that backs
+  // `npx playwright install`. It reads browser revision info from
+  // playwright-core's own package.json, which in the single-file binary is
+  // redirected to the embedded copy via the _resolveFilename hook in
+  // setupEmbeddedDependencies.ts, so the downloaded browser always matches
+  // the bundled runtime exactly.
+  //
+  // The import is lazy (inside this function, not at module top level) so the
+  // _resolveFilename hook is already installed before the registry module loads.
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const { registry, installBrowsersForNpmInstall } = (await import(
+    // @ts-expect-error — internal path, no type declarations
+    "playwright-core/lib/server/registry/index"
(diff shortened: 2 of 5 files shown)


packages/typescript/src/standalone/installPlaywrightBrowsers.ts
```diff
@@ -0,0 +1,175 @@
...
+  if (isSingleFileExecutable()) {
+    // Dynamic import keeps the `bun`-dependent module out of the vitest
+    // environment, which runs under Node and has no `bun` package.
+    const { setupEmbeddedDependencies, getExtractedPlaywrightOopDownloadPath } =
+      await import("./setupEmbeddedDependencies.ts");
+    await setupEmbeddedDependencies();
+    const oopDownloadPath = getExtractedPlaywrightOopDownloadPath();
+    if (!oopDownloadPath) {
+      throw new Error(
+        "Embedded Playwright download worker missing — rebuild the binary.",
```
kossnocorp: It would be better to point to GitHub issues instead. End users won't get that.

packages/typescript/src/mcp/mcpDrivers.ts
```diff
@@ -85,6 +86,7 @@ export async function createPlaywrightDriver(
     logger.debug("Setting extra HTTP headers: {headers}", { headers });
   }
 
+  await ensurePlaywrightChromiumInstalled();
```
kossnocorp, line 89: I'm afraid this won't work on Linux, as it requires `sudo` to install. I will test it.

I think it might be a practical alternative to have a command `alumnium install playwright` and do these shenanigans there. MCP would instruct the agent, which will relay it to the user to run the command.

I would prefer auto-install, but this can cause MCP to timeout, and running MCP server with sudo is mental.

packages/typescript/src/standalone/installPlaywrightBrowsers.ts
```diff
@@ -0,0 +1,175 @@
...
+  const executablePaths = [
+    chromiumExecutable,
+    chromiumHeadlessShellExecutable,
+    ffmpegExecutable,
+  ]
+    .filter(Boolean)
+    .map((exe) => exe?.executablePath())
+    .filter(Boolean) as string[];
+  if (
+    executablePaths.length > 0 &&
```
kossnocorp: An agent spotted? 😉 

```suggestion
    executablePaths.length &&
```

packages/typescript/src/standalone/installPlaywrightBrowsers.ts
```diff
@@ -0,0 +1,175 @@
...
+  // Skip install only if every executable that has an expected path on this
+  // platform is already present on disk.
+  const executablePaths = [
+    chromiumExecutable,
+    chromiumHeadlessShellExecutable,
+    ffmpegExecutable,
+  ]
+    .filter(Boolean)
+    .map((exe) => exe?.executablePath())
+    .filter(Boolean) as string[];
```
kossnocorp:
```suggestion
    .map((exe) => exe?.executablePath())
    .filter(p => p !== undefined)
```

packages/typescript/src/standalone/installPlaywrightBrowsers.ts
```diff
@@ -0,0 +1,175 @@
...
+  if (
+    executablePaths.length > 0 &&
+    executablePaths.every((p) => fs.existsSync(p))
```
kossnocorp: Make it async?

packages/typescript/src/standalone/installPlaywrightBrowsers.ts
```diff
@@ -0,0 +1,175 @@
...
+  ) {
+    logger.debug("Playwright chromium already installed at {executablePath}", {
+      executablePath: executablePaths[0],
+    });
```
kossnocorp: It could be `ffmpegExecutable`, remember you filtered it?

It is better to put them into something like this and check each variable manually, so the paths are stable:

```ts
const executablePaths = [
  chromiumExecutablePath,
  chromiumHeadlessShellExecutablePath,
  ffmpegExecutablePath,
] = await Promise.all([
  locateExePath(chromiumExecutable),
  locateExePath(chromiumHeadlessShellExecutable),
  locateExePath(ffmpegExecutable),
])
```

packages/typescript/src/standalone/installPlaywrightBrowsers.ts
```diff
@@ -0,0 +1,175 @@
...
+    if (!oopDownloadPath) {
+      throw new Error(
+        "Embedded Playwright download worker missing — rebuild the binary.",
+      );
+    }
+    restoreFork = patchForkForOopDownload(oopDownloadPath);
+  }
+
+  try {
+    // ffmpeg is required for video recording (recordVideo is always enabled).
```
kossnocorp, line 101: Maybe make the recording optional?

packages/typescript/src/standalone/installPlaywrightBrowsers.ts
```diff
@@ -0,0 +1,175 @@
...
+    .map((exe) => exe?.executablePath())
+    .filter(Boolean) as string[];
+  if (
+    executablePaths.length > 0 &&
+    executablePaths.every((p) => fs.existsSync(p))
+  ) {
+    logger.debug("Playwright chromium already installed at {executablePath}", {
+      executablePath: executablePaths[0],
+    });
+    return;
```
kossnocorp, line 79: There's a logical error that will not install ffmpeg if the Chromium binary is already installed.

packages/typescript/src/standalone/installPlaywrightBrowsers.ts
```diff
@@ -0,0 +1,175 @@
...
+  const executablesToInstallDeps = [
+    chromiumExecutable,
+    chromiumHeadlessShellExecutable,
+    ffmpegExecutable,
+  ].filter(Boolean);
+  try {
+    await registry.installDeps(executablesToInstallDeps, false);
+  } catch (err) {
+    logger.warn(
+      "Could not install Playwright system dependencies (try running with sudo on Linux): {err}",
```
kossnocorp: > try running with sudo on Linux

Please no 😿

kossnocorp (changes requested):
I like that it auto-installs, but it is better to have `alumnium install playwright` or something similar so it can be run separately (i.e., on CI) and with `sudo` on Linux.

packages/typescript/src/mcp/mcpDrivers.ts
p0deje: `sudo` is only needed to install deps, but that's specifically "best shot" and only captures error if failed.

packages/typescript/src/standalone/installPlaywrightBrowsers.ts
p0deje: I actually write it like this. Ruby taught me that 0 is truthy.

packages/typescript/src/standalone/installPlaywrightBrowsers.ts
p0deje: There is no async version of `existsSync`.

packages/typescript/src/standalone/installPlaywrightBrowsers.ts
p0deje: Maybe it can be done in a follow-up PR.

packages/typescript/src/standalone/installPlaywrightBrowsers.ts
p0deje: Good catch.

packages/typescript/src/standalone/installPlaywrightBrowsers.ts
p0deje: https://github.com/alumnium-hq/alumnium/issues/339

packages/typescript/src/standalone/installPlaywrightBrowsers.ts
p0deje: Fixed, will push shortly.

kossnocorp (approved):
This is fine 🙃

PR #338, opened 2026-05-06
refactor: type-safe, centralized, and validated env vars access

kossnocorp:
This PR wraps `process.env` into type-safe, centralized, and validated API. It utilizes Zod to define schemas for each variable, and centralizes default values and parsing rules.

Now every env variable is parsed, so no unexpected data can come in.

It also adds env vars to logger and prints all the values when the server, MCP, and tests start. Secret values are masked. 

Because the new API is so interconnected, I started getting circular dependency errors, so I enabled a linter rule to disallow circular imports and fixed all the found issues, including many of the existing ones. I had to refactor some classes, like `Model`, `Logger` and `Driver`, and extracted parts of the code into shared modules to break the loops.

### DIFF
--- a/packages/typescript/src/Env.ts
+++ b/packages/typescript/src/Env.ts
@@ -1,0 +1,340 @@
+import ansi from "picocolors";
+import { canonize } from "smolcanon";
+import { xxh32Str } from "smolxxh/str";
+import z from "zod";
+import { Driver } from "./drivers/Driver.ts";
+import { Model } from "./Model.ts";
+import { LoggerSchema } from "./telemetry/LoggerSchema.ts";
+import {
+  arrayString,
+  filenameString,
+  jsonString,
+  pathString,
+} from "./utils/schema.ts";
+import { maskString } from "./utils/string.ts";
+
+export namespace Env {
+  export type VarsRecord = Record<string, unknown>;
+
+  export interface InspectResult {
+    vars: VarsRecord;
+    valid: boolean;
+  }
+}
+
+const secrets = new Set();
+
+let cachedVars: Env.VarsRecord = {};
+let envLogger: LoggerSchema.Like | undefined = undefined;
+
+export const Env = {
+  get ALUMNIUM_CACHE() {
+    return envVar(
+      "ALUMNIUM_CACHE",
+      z
+        .union([z.enum(["sqlite", "filesystem"]), z.stringbool()])
+        .default("filesystem"),
+    );
+  },
+
+  get ALUMNIUM_CACHE_PATH() {
+    return envVar("ALUMNIUM_CACHE_PATH", pathString().optional());
+  },
+
+  get ALUMNIUM_CHANGE_ANALYSIS() {
+    return envVar("ALUMNIUM_CHANGE_ANALYSIS", z.stringbool().default(false));
+  },
+
+  get ALUMNIUM_DELAY() {
+    return envVar("ALUMNIUM_DELAY", z.coerce.number().default(0.5));
+  },
+
+  get ALUMNIUM_EXCLUDE_ATTRIBUTES() {
+    return envVar("ALUMNIUM_EXCLUDE_ATTRIBUTES", arrayString().default([]));
+  },
+
+  get ALUMNIUM_FULL_PAGE_SCREENSHOT() {
+    return envVar(
+      "ALUMNIUM_FULL_PAGE_SCREENSHOT",
+      z.stringbool().default(false),
+    );
+  },
+
+  get ALUMNIUM_LOG_LEVEL() {
+    return envVar("ALUMNIUM_LOG_LEVEL", LoggerSchema.Level);
+  },
+
+  get ALUMNIUM_LOG_PATH() {
+    return envVar("ALUMNIUM_LOG_PATH", pathString().optional());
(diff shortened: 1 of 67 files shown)


packages/typescript/src/server/LlmFactory.ts
```diff
@@ -192,9 +175,9 @@ export class LlmFactory {
   static createAwsLlm(model: Model, cache: BaseCache): BaseChatModel {
     logger.debug(`Creating AWS LLM with model ${model.name}`);
 
-    const accessKeyId = process.env.AWS_ACCESS_KEY ?? "";
-    const secretAccessKey = process.env.AWS_SECRET_KEY ?? "";
-    const region = process.env.AWS_REGION_NAME ?? "us-east-1";
+    const accessKeyId = Env.AWS_ACCESS_KEY ?? "";
+    const secretAccessKey = Env.AWS_SECRET_KEY ?? "";
+    const region = Env.AWS_REGION_NAME ?? "us-east-1";
```
kossnocorp: Prolly need to move defaults to the schemas, also check why `?? ""` are needed.

packages/typescript/src/client/Alumni.ts
```diff
@@ -130,7 +125,7 @@ export class Alumni {
       });
     } else {
       this.client = new NativeClient({
-        model: Model.current,
+        model: Env.ALUMNIUM_MODEL,
```
p0deje: Not related to this PR, but feels like a bug to me:

```suggestion
        model: model ?? Env.ALUMNIUM_MODEL,
```

packages/typescript/src/clients/Client.ts
```diff
@@ -2,7 +2,7 @@ import z from "zod";
 import { AppId } from "../AppId.ts";
 import type { Driver } from "../drivers/Driver.ts";
 import { LlmUsageStats } from "../llm/llmSchema.ts";
-import { Model } from "../Model.ts";
+import { Model } from "../Model/Model.ts";
```
p0deje: I could never understand what the naming pattern is for directories in TypeScript. Should this be similar to `drivers`?

```suggestion
import { Model } from "../models/Model.ts";
```

In fact, it would be great to keep `Model.ts` in the same folder because it creates a 1:1 mapping between Python-TypeScript files, so one doesn't have to think twice.

packages/typescript/src/Model/ModelDev.ts
```diff
@@ -0,0 +1,20 @@
...
+    "google",
+    "deepseek",
+    "meta",
+    "mistralai",
+    "ollama",
+    "xai",
+    "openai",
+  ] as const;
+
+  static Schema = z.enum(ModelDev.devs);
```
p0deje: Why is using `z.enum` + `z.infer` better than just having a basic union type? Is it primarily to have validation when parsing JSON?

packages/typescript/src/Env.ts
p0deje, line 1: Shall we use `z.httpUrl()` for all URLs?

p0deje (approved):

packages/typescript/src/client/Alumni.ts
kossnocorp: @p0deje 100%! Nice catch

packages/typescript/src/clients/Client.ts
kossnocorp: If there were multiple models, i.e., (e.g., `OpenAiModel`, `GoogleModel`) then `models` would make sense. 

In this case, it's `src/Model.ts` (common JS/TS convention is to use the PascalCase for the classes, components, static objects, and their modules) growing and turning its directory `src/Model/`. `src/Model/ModelDev.ts` and `src/Model/ModelProvider.ts`. So the `src/Model/` directory means that everything inside belongs to `Model` but for whatever reason can't be in a single file (too much code or to break circular imports).

So `src/model.py` -> `src/Model/Model.ts` actually maps better to Python than `src/models/Model.ts`, even if it did have semantic sense.

I think 1:1 mapping between Python and TypeScript is a wrong goal, tbh, each language has its conventions, significant module model differences, and context (i.e., client and binary code is heavily mixed in TypeScript). 

---

Regardless, I did the split to address circular imports, but after all the changes, it seems that the split wasn't necessary and it's better to have single file `src/Model.ts` anyway. Haha.

packages/typescript/src/Model/ModelDev.ts
kossnocorp: Unions have no runtime representation, and for Zod to parse it, it needs to have the runtime values.

In general, it's the common approach to replacing native TypeScript enums that aren't supported by Node.js type stripping but have the runtime representation:

```ts
enum Provider {
  OpenAi = "openai",
  Google = "google"
}
```

->

```ts
const providers = ["openai", "google"] as const;

type Provider = typeof providers[number]
```

And unlike TypeScript enums, to pass `Provider` as an argument, you don't need to import it and can use string literal, e.g. `"openai"` whereas the enum would require `Provider.OpenAi`.

packages/typescript/src/Env.ts
kossnocorp: Yeah, good idea!

p0deje (approved):

PR #374, opened 2026-06-06
feat: expand commands in environment variables

p0deje:
Adds support for command substitution in environment variables read by the TypeScript core: when a recognized variable is set to a whole-value `$(command)` (e.g. `OPENAI_API_KEY="$(get-token)`"), Alumnium now runs the command at startup and uses its trimmed stdout as the value, with failures logged (secrets masked) and treated as invalid config so startup aborts. 

Expansion is wired into Env's accessor so it applies to every known variable. Provider API keys (`ANTHROPIC_API_KEY, DEEPSEEK_API_KEY, GOOGLE_API_KEY, MISTRAL_API_KEY, OPENAI_API_KEY, XAI_API_KEY`) were added as `Env` getters and are now passed explicitly into the LangChain clients in LlmFactory so they flow through the same expansion path.

### DIFF
--- a/packages/typescript/src/Env.test.ts
+++ b/packages/typescript/src/Env.test.ts
@@ -1,0 +1,59 @@
+import { afterEach, describe, expect, it, vi } from "vitest";
+import { Env } from "./Env.ts";
+
+afterEach(() => {
+  vi.unstubAllEnvs();
+  Env.reset();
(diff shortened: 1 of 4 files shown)


kossnocorp (approved):
yolo

PR #387, opened 2026-06-14
fix(cache): don't response-cache the actor's action selection

TweedBeetle:
## Problem

The actor agent selects a *mutating* action (click / type / execute-javascript / …) per step. With models that can't run at temperature 0 — notably `gpt-5-nano` (the OpenAI default for the MCP server), whose `temperature` parameter is rejected so it **samples** — the actor occasionally emits a wrong tool (a `ClickTool` where a `TypeTool` was needed), a no-op (no tool call at all), or a faithful-but-corrupted argument (mangled quotes, a dropped `\n`).

`ResponseCache` keys the LLM generation on the exact prompt = (goal + step + accessibility tree, element ids included). So a single bad sample is **written to cache and replayed verbatim on every retry** of the same (goal, step, tree). The action then *never* succeeds on retry — the only escape is changing the prompt, i.e. reloading the page so the element ids change and the cache key no longer matches. This turns a *rare* bad draw into a *permanent* stuck state and makes "just retry" useless.

## Fix

Skip caching the **actor**'s action selection in `ResponseCache.update` (`agentMeta.kind === "actor"`), so each actor invocation re-samples and a retry can escape a bad draw. Read agents (planner / locator / area / retriever) still cache normally; the separate per-element `ElementsCache` layer is unaffected.

## Verification

Against a real, large, already-populated accessibility tree (a multiline textarea on a busy form):

- **Pre-fix:** 8 identical actor invocations → all 8 replayed one frozen no-op draw; only the first was a real API call, the other 7 were instant cache hits (~6 s total). This is the "retries don't work" experience.
- **Post-fix:** the same 8 invocations re-sample → a mix incl. the correct action, recovering by the 2nd invocation (8 real API calls, ~65 s). The wall-clock difference (1 frozen-and-replayed call vs 8 fresh calls) independently confirms the mechanism.

Adds a `ResponseCache` regression test asserting actor-kind generations are not written to cache. `vitest run --project unit` green.

## Note

This is the load-bearing half of a two-part issue. The other half is the actor's per-call mis-selection rate itself (raising `reasoning.effort` and prompt / tool-description hardening were both measured *ineffective* on the dominant populated-field no-op). The cache fix is what makes the failure retry-recoverable rather than permanent.

### DIFF
--- a/packages/typescript/src/server/cache/ResponseCache.test.ts
+++ b/packages/typescript/src/server/cache/ResponseCache.test.ts
@@ -96,6 +96,31 @@ describe("ResponseCache", () => {
     const result = await cache.lookup(prompt1, llmKey);
     expect(result).toBeNull();
   });
+
+  it("does not cache actor action selection so retries re-sample", async () => {
+    const { sessionContext, llmContext, cacheStore, prompt1, llmKey } =
+      await setup();
+    const cache = new ResponseCache(sessionContext, cacheStore, llmContext);
+
+    // Re-key prompt1 with ACTOR metadata (the action-selecting agent).
+    llmContext.clearPromptsMeta([prompt1]);
+    llmContext.assignPromptsMeta([prompt1], {
+      kind: "actor",
+      goal: "fill the field" as never,
+      step: "type into the field" as never,
+      treeXml: "<xml></xml>",
+    });
+
+    const generations = createGenerations("an action");
+    await cache.update(prompt1, llmKey, generations);
+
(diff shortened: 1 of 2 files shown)


p0deje:
I don't think that's a correct solution for your problem. If the prompt returned a no-op operation, or a wrong function call, it's a strong indicator that _the prompt itself needs to be fixed_, typically by providing more details on what exactly to do on the page, or using the learning API to prime the system prompt on how to translate a vague prompt into more concrete steps.

Let's focus on fixing a particular scenario if you have an example in a separate issue.

PR #399, opened 2026-06-22
feat(mcp): read proxy from http_proxy env var in Playwright driver

ursmurli:
## Summary

  - Both the Playwright and Selenium drivers now automatically read proxy settings from http_proxy / HTTP_PROXY / https_proxy / HTTPS_PROXY environment variables — no configuration needed. An explicit proxy object in alumnium:options still takes precedence.
  - Adds proxyFromEnv utility (proxyFromEnv.ts) that parses the env var into a structured proxy config, handling credentials and percent-encoded characters.

  ## Precedence

  1. Explicit alumnium:options.proxy object → highest priority
  2. http_proxy / HTTP_PROXY / https_proxy / HTTPS_PROXY env vars → automatic fallback
  3. No proxy → if neither is set

  ## Test plan

  - [X] `mise run //packages/typescript:types` — passes clean
  - [X] `mise run //packages/typescript:lint` — no new errors (baseline: 3w/15e)
  - [X] `mise run //packages/typescript:format/check` — passes clean
  - [X] `mise run //packages/typescript:test/unit` — 252 passing (pre-existing `mcpDrivers.test.ts` playwright-core failure unchanged)

### DIFF
--- a/packages/typescript/src/mcp/tools/startMcpTool.ts
+++ b/packages/typescript/src/mcp/tools/startMcpTool.ts
@@ -60,9 +60,8 @@ export const startMcpTool = McpTool.define("start", {
             - "permissions" (string[]) — browser permissions to grant, Playwright only, e.g. ["camera"];
             - "planner" (boolean) — enable/disable planner agent;
             - "profile" (string) — name of a persistent browser profile; cookies, sessions, and storage are preserved across restarts in ~/.alumnium/profiles/{name}, e.g. "personal";
-            - "proxy" (object) — HTTP/HTTPS/SOCKS5 proxy, supported for Selenium and Playwright, e.g. {"server": "http://myproxy.com:3128", "bypass": ".com, chromium.org", "username": "usr", "password": "pwd"};
+            - "proxy" (object) — HTTP/HTTPS/SOCKS5 proxy, supported for Selenium and Playwright, e.g. {"server": "http://myproxy.com:3128", "bypass": ".com, chromium.org", "username": "usr", "password": "pwd"}; if omitted, the http_proxy/HTTP_PROXY/https_proxy/HTTPS_PROXY environment variables are used automatically;
             - "recordVideos" (boolean, default true) — record video of the browser session, Playwright only. Can also be disabled via ALUMNIUM_MCP_RECORD_VIDEOS=false;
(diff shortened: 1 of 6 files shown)


packages/typescript/src/mcp/tools/startMcpTool.ts
```diff
@@ -62,6 +62,7 @@ export const startMcpTool = McpTool.define("start", {
             - "profile" (string) — name of a persistent browser profile; cookies, sessions, and storage are preserved across restarts in ~/.alumnium/profiles/{name}, e.g. "personal";
             - "proxy" (object) — HTTP/HTTPS/SOCKS5 proxy, supported for Selenium and Playwright, e.g. {"server": "http://myproxy.com:3128", "bypass": ".com, chromium.org", "username": "usr", "password": "pwd"};
             - "recordVideos" (boolean, default true) — record video of the browser session, Playwright only. Can also be disabled via ALUMNIUM_MCP_RECORD_VIDEOS=false;
+            - "http_proxy" (boolean) — when true, reads proxy settings from the http_proxy/HTTP_PROXY/https_proxy/HTTPS_PROXY environment variables and passes them to Playwright; ignored if "proxy" is also set;
```
p0deje: Let's remove it completely, this should be default behavior

p0deje (commented):
I think we should pick `process.env.http_proxy` automatically without an explicit toggle. This way Alumnium is a good citizen and respects OSes conventions.

packages/typescript/src/mcp/tools/startMcpTool.ts
ursmurli: We retain the HTTP proxy as the default behavior and remove the proxy capability. is that what you mean?

p0deje: 1. By default, if any of the proxy environment variables are present, we parse and use them.
2. If `proxy` is passed in `alumnium:options` - it has more weight and overwrite environment variables.

ursmurli: ok , you mean we don't need a flag here. We by default use HTTP_PROXY url if present.

p0deje (commented):
LGTM, will merge once CI is green.

PR #407, opened 2026-07-09
feat(server): Added support for Cursor as a provider. Defaults to com…

rookieInTraining:
Adds a `cursor` AI provider that runs Alumnium's agents on a Cursor subscription via the official @cursor/sdk local runtime, defaulting to composer-2.5 and authenticated with CURSOR_API_KEY.

Cursor exposes no chat-completions API, so ChatCursor adapts agent runs to a LangChain chat model: a fresh sandboxed throwaway agent per call (falling back gracefully where sandboxing is unsupported), tool calling emulated through a prompt/JSON contract with one corrective re-ask, a withStructuredOutput override that also parses cache hits, and SDK errors mapped to the existing retry semantics. Cursor-specific planner and retriever prompts keep browser-level actions (e.g. tab switching) plannable without tree grounding and single-value retrievals as plain strings.

The compiled binary cannot bundle the SDK's webpack-chunked dist and does not resolve node_modules at runtime, so the provider currently works only with the TypeScript npm package; the binary reports a clear error. Python and Java clients get the enum and default-model entries.


packages/typescript/scripts/build.ts
```diff
@@ -374,7 +374,14 @@ async function main() {
...
           entrypoints: [BIN_SRC_PATH, ...standaloneEmbeddedAssetPaths],
-          external: ["chromium-bidi", "electron"],
+          // NOTE: @cursor/sdk (a transitive dependency of
+          // @alumnium/langchain-cursor) is external because its
+          // webpack-chunked dist loads chunks dynamically
+          // (`require("./" + chunkId + ".js")`), which cannot be bundled into
+          // a single-file executable. Compiled binaries do not resolve
+          // node_modules at runtime either, so ChatCursor reports a clear
+          // unsupported error there.
+          external: ["chromium-bidi", "electron", "@cursor/sdk"],
```
p0deje, line 388: If I read this correctly, it means that a bundled binary cannot use the Cursor SDK. We probably need to embed their JS files directly, similar to how we embed Selenium atoms that are required in in runtime.

packages/typescript/src/llm/ChatCursorCache.test.ts
p0deje, line 1: This should go to `langchain-cursor`?

packages/typescript/src/server/agents/prompts/planner/cursor/system.md
p0deje, line 1: If this prompt is similar to OpenAI, we can just remove it, so that the Cursor model picks up OpenAI defaults.

p0deje: This goes for all other prompts.

packages/typescript/src/server/LlmFactory.ts
```diff
@@ -1,4 +1,5 @@
 import { ChatCodex } from "@alumnium/langchain-codex";
+import { ChatCursor } from "@alumnium/langchain-cursor";
```
p0deje: We'll need to publish it before.

packages/typescript/scripts/build.ts
rookieInTraining: The cursor-sdk [LICENSE](https://github.com/cursor/plugins/blob/main/cursor-sdk/LICENSE) might not allow embedding like Selenium atoms. Any other alternative approach that I can take?

packages/typescript/src/server/agents/prompts/planner/cursor/system.md
rookieInTraining: Noted, will make the change to OpenAI prompt.

packages/typescript/scripts/build.ts
p0deje: MIT allows embedding; this should not be a problem.

packages/typescript/src/llm/ChatCursorCache.test.ts
rookieInTraining: It's present here deliberately as it's consuming the alumnium's `Lchain.ts`. It's basically a compatibility check b/w Lchain and Langchain-cursor

PR #410, opened 2026-07-28
chore(dev): misc dev fixes

kossnocorp:
A few tiny fixes

### DIFF
--- a/mise.toml
+++ b/mise.toml
@@ -1,4 +1,4 @@
-experimental_monorepo_root = true
+monorepo_root = true
 [monorepo]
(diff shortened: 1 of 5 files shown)


mise.toml
```diff
@@ -1,4 +1,4 @@
-experimental_monorepo_root = true
+monorepo_root = true
```
p0deje, line 1: Does this need a new version of mise?

p0deje (approved):

mise.toml
kossnocorp: Well, yeah, you better run `mise self-update -y` regularily anyway

PR #416, opened 2026-08-04
Fix Java Area.get() missing null/noop check

reachsridhard:
# Fix Java `Area.get()` missing null/noop check

## Summary

When the server cannot extract requested data, it returns a `NOOP` result. Java's `Area.get()` was returning `null` instead of the server's explanation string, while every other implementation already handles this correctly.

---

## The Bug

`Area.get()` was calling `result.result().toObject()` directly:

```java
return result.result().toObject();
```

When the server returns a `NOOP` result, `result.result()` is `Data.NoopData.INSTANCE`, and `toObject()` returns `null`. As a result, callers receive `null` instead of the explanation returned by the server.

This behavior is inconsistent with the other SDK implementations, which correctly detect `NOOP` responses and return the explanation string.

---

## The Fix

Check whether the returned value is `null` or a `NOOP` before calling `toObject()`. If so, return the server's explanation wrapped in `Data.StringData`.

```java
Data value = result.result();

if (value == null || value.isNoop()) {
    return new Data.StringData(result.explanation());
}

return value.toObject();
```

This preserves the server-provided explanation instead of silently returning `null`.

---

## Files Changed

| File | Change |
|------|--------|
| `packages/java/src/main/java/ai/alumnium/Area.java` | Added a null/NOOP guard in `get()` and returned `new Data.StringData(result.explanation())` instead of `null`. |

---

## Why This Change?

- Prevents unexpected `null` values from being returned.
- Preserves the server's explanation for `NOOP` responses.
- Aligns the Java implementation with the behavior of the other SDKs.

---

## Test Plan

- Existing tests pass.
- Manually verified that when the server returns a `NOOP` result, `Area.get()` now returns the explanation string instead of `null`.
- Confirmed that normal (non-`NOOP`) responses continue to behave exactly as before.


reachsridhard:
@p0deje 

I’m getting the following CI issue. Could you please check? This should affect all PRs coming from forks, not just this one.

> Can't find 'action.yml' under '.github/actions/setup/post'

PR #421, opened 2026-08-07
chore(ci): set up tests pass threshold on CI + misc fixes

kossnocorp:
This PR introduces `ALUMNIUM_TEST_PASS_THRESHOLD_PCT` to make CI not report failure if the number of test failures is below 5%.

It allows us to notice if the system tests are failing because of the acceptable flakiness % or if there's something more significant.

---

It also includes misc test & types fixes to make CI pass.


.github/workflows/ci.yml
```diff
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - main
+      - ci-pass-threshold
```
p0deje: Let's change to `ci/*` pattern here so we can easily trigger new builds with similarly named branches.

.github/workflows/ci.yml
```diff
@@ -105,9 +111,9 @@ jobs:
     name: Java
     needs: test
     secrets: inherit
-    if: >
-      needs.test.outputs.permitted == 'true'
-        && needs.test.outputs.java-changed == 'true'
+    # if: >
```
p0deje: Don't forget to uncomment.

.github/workflows/ci.yml
```diff
@@ -116,9 +122,9 @@ jobs:
     name: Python
     needs: test
     secrets: inherit
-    if: >
-      needs.test.outputs.permitted == 'true'
-        && (needs.test.outputs.typescript-changed == 'true' || needs.test.outputs.python-changed == 'true')
+    # if: >
```
p0deje: Ditto.

.github/workflows/ci.yml
```diff
@@ -127,9 +133,9 @@ jobs:
     name: TypeScript
     needs: test
     secrets: inherit
-    if: >
-      needs.test.outputs.permitted == 'true'
-        && needs.test.outputs.typescript-changed == 'true'
+    # if: >
```
p0deje: Ditto.

p0deje (approved):

.github/workflows/ci.yml
kossnocorp: Good idea!

p0deje (approved):

PR #433, opened 2026-08-19
feat!(actor): stop clicking on expanded dropdowns

p0deje:
Actor tends to click `<select>` elements to open them first, while it should proceed straight to clicking `<option>` instead. To improve accuracy, select comboboxes now:

1. Don't have `id`.
2. Don't have `focusable` attribute.
3. Have `expanded` attribute.

Same issue happens for other elements (e.g. button) and a related instructions are added to the click tool.

Related evals are added with 0% to 100% accuracy improvement.


kossnocorp (approved):
LGTM

PR #435, opened 2026-08-21
Foundation for waiter script optimization

kossnocorp:
This PR reverts commit caf5e6735f983279aabc170be1233b8bb9ccce52 that I ported in #432 (1deb9a2ddd36fd5673f635b79844e5df42bfe34b).

These two PRs would form a stack and have to be merged together.

### DIFF
--- a/packages/typescript/src/llm/LchainSchema.ts
+++ b/packages/typescript/src/llm/LchainSchema.ts
@@ -230,7 +230,7 @@ export abstract class LchainSchema {
     status: z.union([z.literal("completed"), z.string()]),
     content: z.array(this.MetadataOutput),
     role: z.union([z.literal("assistant"), z.string()]),
-    phase: z.union([z.string(), z.null()]).exactOptional(),
+    phase: z.string().exactOptional(),
   });
 
   static OutputFunctionCall = z.object({
@@ -240,8 +240,6 @@ export abstract class LchainSchema {
     arguments: z.string(),
     call_id: z.string(),
     name: z.string(),
-    caller: z.null().exactOptional(),
-    namespace: z.null().exactOptional(),
   });
 
   static ResponseMetadataOutputItem = z.discriminatedUnion("type", [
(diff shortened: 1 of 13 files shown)


packages/typescript/src/llm/LchainSchema.ts
```diff
@@ -230,7 +230,7 @@ export abstract class LchainSchema {
     status: z.union([z.literal("completed"), z.string()]),
     content: z.array(this.MetadataOutput),
     role: z.union([z.literal("assistant"), z.string()]),
-    phase: z.union([z.string(), z.null()]).exactOptional(),
+    phase: z.string().exactOptional(),
```
p0deje, line 233: This is needed as I'm getting errors otherwise:

```
Error: Failed to serialize generation to stored format:
✖ Invalid input: expected string, received null → at message.data.response_metadata.output[1].phase
```

p0deje: Changing this to nullish suffices btw

```ts
 phase: z.string().nullish(),
```

packages/typescript/src/mcp/tools/startMcpTool.ts
```diff
@@ -76,6 +76,7 @@ export const startMcpTool = McpTool.define("start", {
             - "fullPageScreenshot" (boolean, default false) — capture full-page screenshots.
             - "headers" (object) — extra HTTP headers for every request, supported for Selenium and Playwright, e.g. {"Authorization": "Bearer token"};
             - "headless" (boolean, default false) — run browser headless, supported for Selenium and Playwright;
+            - "newTabTimeout" (number, default 200) — ms to wait for new tab detection, Playwright only;
```
p0deje, line 79: I don't think this option is needed. The new system only blocks if there is a new tab that starts opening but doesn't navigate within a timeout. For such tabs, 200ms is too small and 10s is good enough to wait until the navigation kicks off.

p0deje (commented):
Why revert? This brings back old invalid behavior where every click waits until a new tab opens.

PR #437, opened 2026-08-26
Migrate to AI SDK

kossnocorp:
This PR migrates TypeScript code from LangChain.js to the AI SDK.

The AI SDK is much better typed, has no heavy magic, is better maintained, and is less Pythonic. It will unlock faster iteration on Alumnium agents.


packages/typescript/src/llm/CodexLanguageModel.ts
p0deje, line 1: Would it be better to use https://www.npmjs.com/package/ai-sdk-provider-codex-cli as a follow-up?

packages/typescript/src/llm/LanguageModel.ts
```diff
@@ -0,0 +1,7 @@
+import type {
+  LanguageModelV2,
+  LanguageModelV3,
+  LanguageModelV4,
+} from "@ai-sdk/provider";
+
+export type LanguageModel = LanguageModelV2 | LanguageModelV3 | LanguageModelV4;
```
p0deje, line 7: Do we need to support older models or `LanguageModelV4` would be enough?

packages/typescript/src/server/cache/ChainedCache.ts
p0deje, line 1: I think we only needed because LangChain didn't allow to have multiple caches. Can probably be cleaned up in the follow-up.

packages/typescript/mise.toml
p0deje, line 345: Should this also be removed?

p0deje (approved):

PR #439, opened 2026-08-26
chore: include commit sha in the version

p0deje:
Makes it easier to know which exact commit the current binary is built at:

```
$ dist/bin/alumnium-0.21.0-darwin-arm64 --version
alumnium/0.21.0+1b31a531-dirty darwin-arm64 bun-v24.3.0
```

### DIFF
--- a/packages/typescript/scripts/build.ts
+++ b/packages/typescript/scripts/build.ts
@@ -2,6 +2,7 @@
 
 // This script builds the Alumnium for multiple target platforms using Bun.
 
+import { always } from "alwaysly";
 import { $, type BunPlugin } from "bun";
 import { snakeCase } from "case-anything";
 import fs from "node:fs/promises";
@@ -332,7 +333,9 @@ const standaloneEmbeddedAssetPlugin: BunPlugin = {
 await main();
 
 async function main() {
-  console.log(`🚧 Building Alumnium ${ALUMNIUM_VERSION}...`);
+  const commitSha = await getCommitSha();
+
+  console.log(`🚧 Building Alumnium ${ALUMNIUM_VERSION}+${commitSha}...`);
 
   //#region Clean up
 
@@ -387,6 +390,7 @@ async function main() {
             standaloneEmbeddedAssetPlugin,
           ],
           define: {
+            BUILD_COMMIT_SHA: JSON.stringify(commitSha),
             SINGLE_FILE_EXECUTABLE: "true",
           },
         });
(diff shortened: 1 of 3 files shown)


packages/typescript/scripts/build.ts
```diff
@@ -882,6 +889,23 @@ function getPipModuleName(pkgName: string) {
   return snakeCase(pkgName);
 }
 
+async function getCommitSha(): Promise<string | undefined> {
+  try {
```
kossnocorp: Let's let it fail so we don't break something accidentally and never see the error.

packages/typescript/scripts/build.ts
```diff
@@ -882,6 +889,23 @@ function getPipModuleName(pkgName: string) {
...
+    const sha = (
+      await $`git rev-parse --short HEAD`.cwd(REPO_ROOT_DIR).quiet().text()
+    ).trim();
+    if (!sha) return undefined;
+
+    const changes = (
+      await $`git status --porcelain`.cwd(REPO_ROOT_DIR).quiet().text()
+    ).trim();
```
kossnocorp: `Promise.all` would be quicker, no way `sha` would fail often

packages/typescript/scripts/build.ts
```diff
@@ -882,6 +889,23 @@ function getPipModuleName(pkgName: string) {
   return snakeCase(pkgName);
 }
 
+async function getCommitSha(): Promise<string | undefined> {
+  try {
+    const sha = (
+      await $`git rev-parse --short HEAD`.cwd(REPO_ROOT_DIR).quiet().text()
+    ).trim();
+    if (!sha) return undefined;
```
kossnocorp: `always(sha)` would be better, `undefined` here is _undefined_ behavior.

packages/typescript/scripts/build.ts
```diff
@@ -388,6 +392,9 @@ async function main() {
           ],
           define: {
             SINGLE_FILE_EXECUTABLE: "true",
+            ...(commitSha
+              ? { BUILD_COMMIT_SHA: JSON.stringify(commitSha) }
+              : {}),
```
kossnocorp: If we were to remove all the defensive code, this condition would simplify to:

```suggestion
            BUILD_COMMIT_SHA: JSON.stringify(commitSha),
```

kossnocorp (changes requested):
I feel nitpicky

kossnocorp (approved):
Good to go, great idea btw!

PR #447, opened 2026-08-31
feat(mcp): domain allowlist/denylist SSRF protection for do() navigation

ursmurli:
## Summary
- Adds a `NavigationPolicy` enforced on every driver to protect `do()` navigation against SSRF-style abuse (typed URL, clicked link, JS redirect, back/forward, new tab).
- **Always-on by default, no opt-in/opt-out**: cloud metadata/link-local IPs (`169.254.*`), IPv6 link-local (`fe80::`), and `file://` are blocked on every session unconditionally.
- New MCP `start` options `alumnium:options.allowlistDomains` / `denylistDomains`:
  - `denylistDomains` alone (no allowlist) blocks specific extra hosts while leaving everything else — including `localhost`/`127.0.0.1` — open.
  - `allowlistDomains` locks a session down to an approved domain set (default-deny), which also then denies loopback/localhost unless explicitly re-added.
  - An allowlist match always overrides a denylist match, so a caller can carve out one host from under a broader deny wildcard.
- Closes the CVE-2026-49857 bypass class: an IPv4-mapped IPv6 literal (`::ffff:127.0.0.1`, normalized by the URL parser to hex `::ffff:7f00:1`) is unwrapped back to dotted-decimal before matching, so decimal-based patterns still catch it.
- Enforced at two points so every navigation vector is covered: pre-check inside each driver's `visit()`, and a post-invoke check in `BaseTool.executeToolCall` (catches clicks, redirects, tab switches, back/forward) — not just the explicit "go to URL" tool.

## Test plan
- [ ] `mise //packages/typescript:test/unit`
- [ ] `mise //packages/typescript:types`
- [ ] `mise //packages/typescript:lint`
- [ ] Manual: `start` with no `alumnium:options` → `do("open http://169.254.169.254/")` blocked, `do("open http://localhost:3000/")` allowed.
- [ ] Manual: `start` with `allowlistDomains` set → navigation outside the allowlist (including `localhost`) is blocked.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


packages/typescript/src/tools/BaseTool.ts
```diff
@@ -26,6 +26,8 @@ export abstract class BaseTool {
     const tool = new ToolClass(toolArgs);
     await tool.invoke(driver);
 
+    driver.navigationPolicy.check(await driver.url());
```
p0deje: Tools should not be checking for navigation policy as long as we only constraint driver's `visit`.

packages/typescript/src/mcp/tools/startMcpTool.ts
```diff
@@ -66,9 +67,11 @@ export const startMcpTool = McpTool.define("start", {
           Top-level options:
 
           Alumnium-specific options go in "alumnium:options":
+            - "allowlistDomains" (string[]) — regex patterns (case-insensitive, matched against the hostname or full URL) of domains do() is allowed to navigate to. When non-empty, activates full lockdown: every navigation (explicit navigate, links, redirects, back/forward, tab switches) is checked, and a URL that doesn't match is blocked (default-deny), including loopback addresses (localhost/127.0.0.1/[::1]/0.0.0.0). When omitted, sessions are still protected by Alumnium's always-on denylist below — this only adds the stricter default-deny gate;
```
p0deje: This gets added to MCP tools context of every agent, so let's keep it concise:


```suggestion
            - "allowlistDomains" (string[]) — regex patterns matched against the URL. When non-empty, do() blocks any navigation that doesn't match the patterns.
```

packages/typescript/src/mcp/tools/startMcpTool.ts
```diff
@@ -66,9 +67,11 @@ export const startMcpTool = McpTool.define("start", {
...
             - "autoswitchToNewTab" (boolean, default true) — auto-switch to newly opened tabs;
             - "baseUrl" (string) — URL to navigate to automatically after driver start, e.g. "https://example.com";
             - "changeAnalysis" (boolean, default true) — enable UI changes analysis agent;
+            - "denylistDomains" (string[]) — additional regex/literal patterns to block, merged with Alumnium's built-in SSRF-protection denylist and always enforced, even without "allowlistDomains". The built-in denylist always blocks cloud metadata/link-local IPs (e.g. 169.254.169.254), IPv6 link-local, and file:// — this cannot be disabled. Loopback addresses (localhost/127.0.0.1/[::1]/0.0.0.0) are only added to the denylist once "allowlistDomains" is set (otherwise they're left open, since they're the common target for testing a locally-running app). When "allowlistDomains" is set, a match there always overrides a "denylistDomains" match;
```
p0deje: This gets added to MCP tools context of every agent, so let's keep it concise:


```suggestion
            - "denylistDomains" (string[]) — regex patterns matched against the URL. When non-empty, do() blocks any navigation that matches the patterns.
```

p0deje (changes requested):

packages/typescript/src/tools/BaseTool.ts
ursmurli: Visit is not the only threat vector here, there are other way non-visit tools like click or some press key which triggers a redirect , an in-page js redirect, switch to a next tab , backward , forward . 

This particular line will make this navigation policy is not subverted regardless of what tool is used or what server side or client side redirects do.

p0deje (approved):

p0deje (approved):

PR #448, opened 2026-08-31
test(mcp): add system tests for NavigationPolicy wiring

ursmurli:
## Summary
Stacked on #447 — adds a real-browser (Playwright + Selenium) system-test suite proving `NavigationPolicy`'s enforcement wiring holds end-to-end, on top of the exhaustive pure-logic unit tests already in `NavigationPolicy.test.ts`.

New file: `packages/typescript/tests/system/navigationPolicy.test.ts`, plus one new fixture `packages/python/examples/support/pages/navigation_target_page.html` (parametrized via `?to=<url>`, reused for every click/redirect/new-tab case).

Covers both enforcement hooks against real browsers:
- **Hook #1** (`driver.visit()`, pre-navigation): open mode (arbitrary domain / loopback allowed, always-on denylist for cloud metadata / `file://` / IPv6 link-local / IPv4-mapped-IPv6 metadata blocked, mapped-loopback allowed), lockdown mode (allowlisted hostname allowed, loopback denylist + default-deny blocked), and `allowedFilePaths` (allowed prefix succeeds, outside-prefix and `..`-traversal blocked).
- **Hook #2** (`BaseTool.executeToolCall`, post-invoke): a real click that JS-redirects to a blocked target (open + lockdown mode), a real click that opens a new tab to a blocked/allowed target, confirming the click's side effect already ran but the subsequent check still throws.

Design notes (see plan for full rationale):
- The exhaustive rule matrix (every allow/deny/bypass permutation) intentionally stays unit-tested only — these system tests are a representative subset proving the *wiring*, not re-deriving the policy logic, to avoid CI flakiness from real-network/socket edge cases with no added confidence.
- Uses `https://example.com` (RFC 2606, IANA-reserved, content stable by design) as the one real external domain, consistent with this repo's existing pattern of hitting real public sites in system tests (`the-internet.herokuapp.com`, `seleniumbase.io`, etc.).
- Appium is explicitly skipped for this suite (scope: Playwright + Selenium only), even though `AppiumDriver.visit()` wires the identical pre-nav hook.

## Test plan
- [ ] CI: `TypeScript / System Tests (playwright, *)` and `(selenium, *)` pick up and pass the new suite (no CI wiring needed — glob-discovered).
- [ ] `mise //packages/typescript:types` / `:lint` / `:format` clean (already verified locally).
- [ ] Manually reviewed: hardcoded case expectations verified against `NavigationPolicy.evaluate()` directly (standalone script, all cases match).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/python/examples/support/pages/navigation_target_page.html
+++ b/packages/python/examples/support/pages/navigation_target_page.html
@@ -1,0 +1,27 @@
+<!doctype html>
+<html>
+  <head>
+    <title>Navigation Target Page</title>
+  </head>
+  <body>
+    <h1>Navigation Target Page</h1>
+    <p>Used to test the navigation domain policy's post-invoke hook.</p>
+
+    <button id="redirect-btn" onclick="redirect()">Redirect Here</button>
+    <button id="open-tab-btn" onclick="openTab()">Open Tab Here</button>
+
+    <script>
+      function target() {
+        return new URL(location.href).searchParams.get("to");
+      }
+
+      function redirect() {
+        location.href = target();
+      }
+
+      function openTab() {
+        window.open(target(), "_blank");
+      }
+    </script>
+  </body>
+</html>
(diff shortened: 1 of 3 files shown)


ursmurli:
Merged directly into #447's branch (`feat/add-domain-allowlist-mcp`) instead of stacking — closing this PR since its commits now live there.

PR #455, opened 2026-09-10
feat(mcp): agentic and direct modes

p0deje:
Adds two MCP execution modes: 
- **agentic** (default), where Alumnium uses an LLM to interpret goals through `do`, `get`, `check`, similar to the currently released MCP server. This mode is _slower_ since extra LLM calls are made, but _cheaper_ as the coding agent burns fewer tokens. It also supports caching.
- **direct**, which exposes actor actions and accessibility-tree retrieval so the coding agent controls execution without Alumnium making LLM calls. This mode is _faster_ since, but _more expensive_ as the coding agent needs to pull more context.

Mode can be selected with the `--mode agentic|direct` flag or `ALUMNIUM_MCP_MODE`.
Both modes share session setup and cleanup across supported browsers and mobile platforms.


packages/typescript/src/mcp/McpServer.ts
```diff
@@ -66,12 +79,8 @@ export class McpServer {
     });
   }
 
-  /**
-   * Run the MCP server using stdio transport.
-   */
-  async run(): Promise<void> {
+  async run(transport: Transport = new StdioServerTransport()): Promise<void> {
     logger.info("Starting MCP server with stdio transport");
-    const transport = new StdioServerTransport();
```
kossnocorp, line 74: Keep it here

packages/typescript/src/mcp/McpServer.ts
```diff
@@ -28,23 +32,31 @@ const MCP_TOOLS = [
...
+];
+
 /**
  * MCP Server that wraps Alumnium functionality for AI agents.
  */
 export class McpServer {
   #server: Server;
 
-  constructor() {
+  constructor({ mode = "agentic" }: { mode?: McpMode } = {}) {
```
kossnocorp: Pls define `namespace McpServer { ...` with `interface Props { mode: ...`

packages/typescript/src/mcp/McpCommand.ts
```diff
@@ -11,13 +13,20 @@ export const McpCommand = CliCommand.define({
...
+    }),
+  }),
 
-  action: async ({ logFilenameHint }) => {
+  action: async ({ args, logFilenameHint }) => {
     Logger.path = { filename: logFilenameHint };
     await Logger.initEnv(logger);
 
-    const server = new McpServer();
+    const server = new McpServer({ mode: args.mode ?? Env.ALUMNIUM_MCP_MODE });
```
kossnocorp:
```suggestion
    const server = new McpServer({ mode: args.mode || Env.ALUMNIUM_MCP_MODE });
```

kossnocorp (approved):

